PR #9, opened 2014-05-20
XEP-0297: Stanza Forwarding

jaragont:
XEP-0297: Stanza Forwarding


jlaine:
I don't believe this was ever merged. It lacks documentation, and I'm not sure what the point of the QSharedPointer is. Also, is it doesn't seem possible to clear the forwarded message?

LNJ2:
At least there is https://github.com/qxmpp-project/qxmpp/blob/master/src/client/QXmppCarbonManager.h for carbons.

olesalscheider:
It accesses the forwarded dom element internally and just returns the forwarded message.
In contrast, this PR exposes XEP-0297 to the user.

jlaine:
OK, I'm not sure there is any value in exposing XEP-0297 directly since we already have a higher-level manager, so I'm going to close this.

PR #17, opened 2014-09-12
FIX:Add handle non-sasl authentication error(not-authorized) response

KangLin:
FIX: The signal vCardReceived() is emited when only getting vCards of entries in Roster.


KangLin:
Add handle non-sasl authentication error(not-authorized) response

jlaine:
OK this pull request is now just a mess, with a whole bunch of unrelated commits. I am closing it, please reopen one pull request per feature / fix

PR #26, opened 2015-01-06
Update package version

poxip:
`pkg-config --modversion qxmpp` gives 0.8, but version is 0.8.1 (this can be very confusing, because of some crashes on 0.8)


jlaine:
Thanks for spotting this, I had indeed forgotten to update the version. I can't silently "re-do" an 0.8.1 release, I have pushed an update to 0.8.2, which I will tag later today.

PR #31, opened 2015-01-20
protect against self-assignment

longwei:


0xd34df00d:
Great, thanks for your work!

Won't you mind slightly changing the code style please, though? It would be great if each `return` would be on next line after corresponding `if`.

jlaine:
I may be missing the point here, but for all the cases where we are purely assigning the d-pointer, why do we need to do anything at all?

longwei:
@jlaine semantic allow self-assignment could cause subtle memory corruption issue.
purely assign d-pointer imply:
1. change which object d pointer point to
2. increment ref count of this data object
3. and decrease ref count of that data object, delete it if 0;

assigning d-pointer didn't get around with it, it is the assign d-pointer that point to a private implementation that /*happened to*/ handle self self assignment save us here. But private class /*may*/ not.

anyway I think it is a good idea to add this protection to have a peace of mind. :)
My two cents.

jlaine:
This change feels wrong, I have browsed through the Qt codebase and checking for self assignment is not a pattern there. There may be *some* isolated places where we need to do this, but I'd like a testcase illustrating the problem first.

longwei:
@jlaine yep, I see you point. current usage are all QSharedDataPointer/QShareData pair. I am closing this PR

PR #43, opened 2015-02-25
fix issue with connecting to Google Talk service

vogel:
My Google Talk account is vogel@kadu.im
It used domain kadu.im but conencts to host talk.google.com (custom
host/port settings).

When I try to connect with setIgnoreSslErrors(false) I get
QAbstractSocket::SslHandshakeFailedError error. This fixes that issue.

### DIFF
--- a/src/client/QXmppOutgoingClient.cpp
+++ b/src/client/QXmppOutgoingClient.cpp
@@ -382,72 +382,75 @@ void QXmppOutgoingClient::handleStanza(const QDomElement &nodeRecv)
         }
 
         // handle authentication
-        const bool nonSaslAvailable = features.nonSaslAuthMode() != QXmppStreamFeatures::Disabled;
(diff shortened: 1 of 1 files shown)


jlaine:
Also could you please answer the question I asked by email: why are you forcing the host name? The nominal way to connect is to do DNS SRV lookups based on the JID domain.

vogel:
I've not received any questions by e-mail. Maybe it got into spam folder or I just missed the e-mail.

And you are right - the forcing is not needed. It was in my configuration for several years (I'm not the first XMPP developer for Kadu, so there is a lot of cargo programming).

vogel:
So that mean I will just have to add option to setIgnoreSslErrors(true) if certificate does not match. I would much more prefer to use setCaCertificates() and to allow users to accept certificates, but I do not see how I could get non-verified certificate from error() signal of QXmppClient.

vogel:
I'll find another way.

PR #44, opened 2015-03-06
only authenticate when credentials are provided

vogel:
It helps with registration of new accounts:
http://xmpp.org/extensions/xep-0077.html

I'm not sure if this is the best way to do it, but it definitely works

### DIFF
--- a/src/client/QXmppOutgoingClient.cpp
+++ b/src/client/QXmppOutgoingClient.cpp
@@ -382,72 +382,75 @@ void QXmppOutgoingClient::handleStanza(const QDomElement &nodeRecv)
         }
 
         // handle authentication
-        const bool nonSaslAvailable = features.nonSaslAuthMode() != QXmppStreamFeatures::Disabled;
-        const bool saslAvailable = !features.authMechanisms().isEmpty();
-        if (saslAvailable && configuration().useSASLAuthentication())
+        if (!configuration().user().isEmpty() || !configuration().password().isEmpty())
         {
-            // supported and preferred SASL auth mechanisms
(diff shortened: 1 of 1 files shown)


vogel:
Do you want me to change that in any way?

jlaine:
I need to give this some more thought, I cannot remember what state this leaves the client in: what signal will you use to know you can proceed with registration?

In the meantime you can look into writing some tests for this.

vogel:
It emits "connected" properly. You can see here: https://gitlab.com/kadu/kadu/blob/master/plugins/jabber_protocol/services/jabber-register-account.cpp

I'll look what I can do about unit tests tommorow.

0xd34df00d:
@vogel I managed to implement in-band registration without any modifications to the current core code.

For example, see
https://github.com/0xd34df00d/leechcraft/blob/master/src/plugins/azoth/plugins/xoox/inbandaccountregsecondpage.cpp (note how `Client_` is used) and https://github.com/0xd34df00d/leechcraft/blob/master/src/plugins/azoth/plugins/xoox/regformhandlerwidget.cpp

vogel:
Thanks. I'll check it - it is better not to rely on unpatched/unupgraded version of library.

vogel:
@0xd34df00d Wow it works. I've added these two lines to my configuration:

configuration.setUseNonSASLAuthentication(false);
configuration.setUseSASLAuthentication(false);

I'm not sure if this is recommended way to do what I want, but if so, this pull request can be ignored.

Anyway I'm happy that I won't have to wait to new QXmpp release to release new Kadu. Thanks again!

vogel:
Closing.

PR #49, opened 2015-04-19
Allow to set the install paths separately

olesalscheider:
This patch allows to override the following variables:
HEADER_INSTALL_PATH: Install path for the header files
LIB_INSTALL_PATH: Install path for the libraries
DOC_INSTALL_PATH: Install path for the documentation

If these variables are not set they are assigned the old values.

This is necessary for e. g. Exherbo, where docs go to /usr/share but
libraries and headers go to /usr/arch-triplet/{lib,include}.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ SUPPORTED PLATFORMS
 It should work on all the plaforms supported by Qt. For a complete list of
 platforms support by Qt, see:
 
-http://qt-project.org/doc/supported-platforms.html
+https://doc.qt.io/qt-5/supported-platforms.html
 
 In past, we have tested on variety of platforms:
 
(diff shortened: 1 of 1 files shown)


jlaine:
If you are going to override paths during packaging, why not just move the files to the expected location?

olesalscheider:
Exherbo (which is a source based distribution) has a policy that copying files or patching is only acceptable if upstream does not accept a patch.
This really makes sense for projects with build systems that have standardised ways to override the install paths (e. g. automake). In this case, the paths do not have to be set per package, but only once.

But since qxmpp uses qmake and requires to set the paths in the package anyway, I would not mind that much if you decide not to accept this patch.

PR #51, opened 2015-05-11
XEP-0055 Jabber Search, XEP-0004 Multiple Items in Form Results

LightZam:
1. XEP-0055 protocol (http://www.xmpp.org/extensions/xep-0055.html)
2. XEP-0004 implement for XEP-0055 (http://xmpp.org/extensions/xep-0004.html#protocol-results)
3. refactor XEP-0004 for reusability
4. unit test for XEP-0055 
5. unit test for XEP-0004 multiple items in form results

### DIFF
--- a/src/base/QXmppPubSubIq.cpp
+++ b/src/base/QXmppPubSubIq.cpp
@@ -208,6 +208,7 @@ void QXmppPubSubIq::parseElementFromChild(const QDomElement &element)
     {
     case QXmppPubSubIq::ItemsQuery:
     case QXmppPubSubIq::PublishQuery:
+    case QXmppPubSubIq::RetractQuery:
         childElement = queryElement.firstChildElement("item");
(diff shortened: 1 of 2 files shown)


LightZam:
Any suggestion to this pull request?
Will it has chance to be merged?

LightZam:
okay, i am start to understand qxmpp's PR style, i will be more friendly at next time.

yes, i can. but where is the document location that you expect to be? and the format

jlaine:
The comments should be in the .cpp file, they will automatically be extracted by Doxygen, for example:

https://github.com/qxmpp-project/qxmpp/blob/master/src/client/QXmppClient.cpp#L234

To rebuild the docs locally, install "doxygen", then:

qmake QXMPP_USE_DOXYGEN=1
make
make docs

=> look in doc/html

PR #53, opened 2015-05-15
fix issue#52

LightZam:
use XEP-0203 instead of XEP-0091, if there have XEP-0203.


jlaine:
Can you please add a unit test to lock down this behaviour and a comment about why we parse XEP-0203 second in the block. The idea is that your change does not get lost in some future refactoring.

LightZam:
okay, i will do it, thanks the advice.

PR #58, opened 2015-06-03
[update] fix RetractQuery [new] unit test for RetractQuery 

LightZam:
and unit test for it


jlaine:
Can we please have just one change (and its test) per PR, this make it much easier to review

LightZam:
yes, sure. this was my mistake, i will remove DeleteQuery part.

jlaine:
OK, next time can you squash your commits, so that there is a single commit for your change? Otherwise the history looks rather messy.

http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit

PR #76, opened 2015-09-22
Added possibility to QXmppServer to route a stanza to a specific resource

### LINKED ISSUE #7: QXmppServer: Route stanza for a bare jid to specific resource
As far as I can see, it isn't possible to tell the QXmppServer to route a stanza with a bare JID in the 'to' attribute to a specific resource without changing the 'to' attribute.

If I want to route a message only to the resource with the highest presence priority or the most recently used resource, I must not change the 'to' attribute as stated in RFC 6121 chapter 8.5.2.1.1 (Server Rules for Processing XML Stanzas / Local User / Avalable or Connected Resource / Message) last paragraph: 

```
In all cases, the server MUST NOT rewrite the 'to' attribute (i.e.,
it MUST leave it as <localpart@domainpart> rather than change it to
<localpart@domainpart/resourcepart>).
```

So this cannot be achieved with QXmppServer. I think that could easily be achieved by either adding a 2nd Parameter const QString& resource = QString() to sendElement(...) and sendPacket(...)

```
bool sendElement(const QDomElement &element, const QString& resource = QString());
bool sendPacket(const QXmppStanza &stanza, const QString& resource = QString());
```

 or by providing overloaded methods

```
bool sendElement(const QDomElement &element);
bool sendElement(const QString& resource, const QDomElement &element);
bool sendPacket(const QXmppStanza &stanza);
bool sendPacket(const QString& resource, const QXmppStanza &stanza);
```

As I don't know the ideas / architecture / coding rules behind qxmpp I cannot estimate whether any of the suggestions would fit or whether it could cause problems.

Currently I think it would be enough to only add the possibility to just specify the resource but maybe someone can think of a use case for specifying the full JID.

servonic:
Added possibility to QXmppServer to route a stanza to a specific resource without the need to add the resource to the stanza's JID. Closes #7


servonic:
I updated the pull request (I hope I understood you correctly):
I moved the if into routeData (passing the resource through sendElement / sendPacket) and prevent routing if "to" is already a full JID.
Please let me know whether you want me to sqash it into a single commit, or if you still have concerns whether you have a better idea to solve this.

servonic:
@jlaine I rebased (and tested again) the changed I did to the pull request. Unfortunately the rebasing seems to have deleted our comments - sorry for that!
I think I moved the decision to the point you wanted it. 
Can you please recheck and if you still see problems, please let me know what I should change.

PR #79, opened 2015-09-29
QXmppTransferManager: Fixed device management

Kaffeine:
Prevent device reparenting and close the dev only if we opened it.

sendFile() method expects the device to be opened, but
QXmppTransferJob::terminate() closes the dev unconditionaly, which
breaks reusable QIODevice workflow.

E.g. current code closes an input-output buffer after transfer,
before the data from buffer read.

Device parent setup also cause the device deletion, which is not
expected in some cases.

This issue affects https://github.com/TelepathyQt/telepathy-nonsense project, which uses the same device to read/write file content from/to xmpp and to write/read the content to/from TelepathyQt FileTransfer interface.

I understand that this can lead to memory-leak, if a device, passed to QXmpp is allocated on heap, have no parent and not deleted manually, but I'm sure that this is the right thing to do. (Better, than to close and delete passed device)
We should close and delete only device, which we constructed and opened.

### DIFF
--- a/src/client/QXmppTransferManager.cpp
+++ b/src/client/QXmppTransferManager.cpp
@@ -1307,7 +1307,7 @@ QXmppTransferJob *QXmppTransferManager::sendFile(const QString &jid, const QStri
     fileInfo.setDescription(description);
 
     // open file
-    QIODevice *device = new QFile(filePath);
+    QIODevice *device = new QFile(filePath, this);
     if (!device->open(QIODevice::ReadOnly))
     {
         warning(QString("Could not read from %1").arg(filePath));
@@ -1342,6 +1342,7 @@ QXmppTransferJob *QXmppTransferManager::sendFile(const QString &jid, const QStri
 /// Returns 0 if the \a jid is not valid.
 ///
 /// \note The recipient's \a jid must be a full JID with a resource, for instance "user@host/resource".
+/// \note The ownership of the \a device should be managed by the caller.
 ///
 
 QXmppTransferJob *QXmppTransferManager::sendFile(const QString &jid, QIODevice *device, const QXmppTransferFileInfo &fileInfo, const QString &sid)
@@ -1361,8 +1362,6 @@ QXmppTransferJob *QXmppTransferManager::sendFile(const QString &jid, QIODevice *
         job->d->sid = sid;
     job->d->fileInfo = fileInfo;
     job->d->iodevice = device;
-    if (device)
-        device->setParent(job);
 
     // check file is open
     if (!device || !device->isReadable())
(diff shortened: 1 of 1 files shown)


olesalscheider:
+1
I can confirm the issue and the patches fix it.

src/client/QXmppTransferManager.cpp
```diff
@@ -1361,8 +1361,6 @@ QXmppTransferJob *QXmppTransferManager::sendFile(const QString &jid, QIODevice *
         job->d->sid = sid;
     job->d->fileInfo = fileInfo;
     job->d->iodevice = device;
-    if (device)
```
0xd34df00d: Probably a better way is to check the device parent and reparent it only if its parent is `nullptr`? Seems like a sane and expected behavior to me.

src/client/QXmppTransferManager.cpp
```diff
@@ -426,7 +426,7 @@ void QXmppTransferJob::terminate(QXmppTransferJob::Error cause)
     d->state = FinishedState;
 
     // close IO device
-    if (d->iodevice)
+    if (d->iodevice && d->iodevice->parent() == this)
```
0xd34df00d: I don't like the idea of "having a parent equal to `this`" being a sign of "we have opened this device". Perhaps the `d`-implementation could be modified to have this flag explicitly?

jlaine: I agree with Georg, I'd rather see an explicit boolean saying "we created this device"

Kaffeine:
@0xd34df00d, @jlaine I fixed mentioned objections, can you please re-review this PR?

jlaine:
Can you please document the new behaviour in the sendFile(QIODevice*) method, namely that ownership of the QIODevice should be managed by the caller?

Kaffeine:
Sure.

jlaine:
Actually rereading the patches the ownership model is confusing: we take ownership of the QIODevice except if it already had a parent? I don't really have any similar Qt APIs in mind, usually they take or don't take ownership, but not conditionally. Any thoughts on this?

Kaffeine:
In my opinion we should not take the ownership at all (and this is what we had in the first PR version), but I followed

>0xd34df00d
>Probably a better way is to check the device parent and reparent it only if its parent is nullptr? Seems like a sane and expected behavior to me.

jlaine:
I would agree we should not take ownership, and document it as such. If you decide to use a QIODevice (as opposed to just passing a file path), you have elected to use a low-level API. As such, you shoud manage the lifetime of your QIODevice.

Kaffeine:
@jlaine: Done. (The docs seems to be not too wordy, so I think the one-line note would be enough?)

0xd34df00d:
@jlaine agreed on such behavior.
@Kaffeine Yep, something along the lines of `The caller remains the owner
of the device.`.

Kaffeine:
@jlaine , @0xd34df00d ping. (please re-review and may be merge updated commits)

Kaffeine:
Thanks!

PR #85, opened 2015-10-21
File transfer: Make sure the filename is being converted to a QUrl correctly

fbeutel:
As it is already done in the QXmppTransferJob::accept method, QXmppTransferManager::sendFile should also convert the local file path to a QUrl using the QUrl::fromLocalFile method.


jlaine:
Can you please add a unit test for this?

fbeutel:
Sorry, should have done that right away ;)
I added one that demonstrates the problem with the original implementation.

PR #86, opened 2015-10-23
Implemented Client State Indication (XEP-0352)

fbeutel:


fbeutel:
Oh, sorry, should have rebased to original master first...

PR #87, opened 2015-10-23
Implemented xep-0352: Client State Indication

fbeutel:


src/base/QXmppStreamFeatures.cpp
```diff
@@ -94,6 +94,16 @@ void QXmppStreamFeatures::setTlsMode(QXmppStreamFeatures::Mode mode)
     m_tlsMode = mode;
 }
```
jlaine: Please document these two members

src/client/QXmppClient.cpp
```diff
@@ -320,6 +324,27 @@ bool QXmppClient::isConnected() const
     return d->stream->isConnected();
 }
 
+/// Returns true if the current client state is "active", false if it is "inactive"
+/// See xep-0352.
+
+bool QXmppClient::clientState()
```
jlaine: I am not sure about the member name, or using a boolean to convey this information. Code like this does not seem very readable:

if (client->clientState()) {
..
}

renini:
+1 will this get merged into master?

jlaine:
Superseed by PR #159

Issue #100, opened 2016-06-16
Add a .prf "feature file" for QMake for simpler builds


jankusanagi:
Apparently, QXMPP doesn't include a .prf file, which other Qt-based libraries use to allow easy (and safe) "feature inclusion" when creating makefiles with qmake, from a .pro file.

If a library provides such a feature file, an application which uses it, and uses qmake as buildsystem, can just have a line like "CONFIG += qxmpp" (whatever the name of the .prf file is), and then, there's no need to manually link to qxmpp with something like "LIBS += -lqxmpp", and the include files are guaranteed to be in the include path.

The main advantage is the assurance that the library to link will be correct, regardless of the actual name the library file has in the system (like libqxmpp-qt5, or whatever a specific distribution ends up calling it).

If a GNU/linux distribution wants to use a different name for some reason, or provide both qt4 and qt5 co-installable builds, they can tune this .prf file, and the apps depending on QXMPP don't need to worry about this.

Samples:
**QOAuth**: https://github.com/ayoy/qoauth/blob/master/oauth.prf
**QCA**: https://quickgit.kde.org/?p=qca.git&a=blob&h=f25440f244ad0249d9a29df77a4f81949a277542&f=crypto.prf.cmake&o=plain
(I don't know why this one is called .prf.cmake, but the result is just crypto.prf)

You can see on their README their instructions to start using QOAuth:
https://github.com/ayoy/qoauth/blob/master/README#L31

Thanks!!

EDIT: similar to what's provided in the qxmpp.pc file that's generated when qxmpp is built.

jlaine:
Sounds good to me, would you mind providing a pull request?

jankusanagi:
Great!

I don't mind providing a pull request, although I don't think I'm really qualified for this xD

I've prepared a file that seems to work for my setup, based on the one from QOAuth. I'll provide a PR in a moment, but it will probably need some checking from someone more "in the know" about this. Especially the part targeting other non-GNU OSes :p

jankusanagi:
Done!

https://github.com/qxmpp-project/qxmpp/pull/101

PR #101, opened 2016-06-27
Added a qmake CONFIG feature file

jankusanagi:
.prf file provided as requested in https://github.com/qxmpp-project/qxmpp/issues/100

It's a rough adaptation of the feature file provided by the QtOAuth library. There are things I don't really understand, and conditionals for platforms I don't use, but it can be used as a starting point.

It will certainly make life easier (and safer) for people creating qmake-based projects which link to QXMPP, especially considering how some distributions might provide both Qt4 and Qt5 builds of QXMPP with different sonames and such.

I've tested it on my system, by copying the new file to /usr/lib/qt{4|5}/mkspecs/features/ and using "CONFIG += qxmpp" on my .pro file, and it works as expected.

Cheers!

### DIFF
--- a/qxmpp.prf
+++ b/qxmpp.prf
@@ -1,0 +1,46 @@
+QXMPP_INCDIR = $$[QT_INSTALL_HEADERS]
+QXMPP_LIBDIR = $$[QT_INSTALL_LIBS]
+
+## Distributions can adjust this for Qt 4
+QXMPP_INCDIR ~= s!/qt5*!!
+QXMPP_LIBDIR ~= s!/qt5*!!
+
+CONFIG *= qt
+
+
+# if we are including qxmpp.prf from the qxmpp tree (and not utilizing it as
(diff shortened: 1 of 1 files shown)


Kaffeine:
@jlaine , can you please restart travis-ci build for this PR?
Even if you would not accept it, it would be better for everyone to know does it break build or not.

As far, as I see, this change do not affect build anyhow, so there should be a green light.

jankusanagi:
I don't know about the CI checks, but right above this comment box it states "This branch has no conflicts with the base branch" and a nice green check icon...

jankusanagi:
Apparently pulling from your master branch to my clone helped the CI checks... all green now.

jankusanagi:
Hmm... ping? =)

jlaine:
Thanks for your patch! Should we be modifying our .pro file to install this?

Issue #96, opened 2016-03-15
Handle IQ responses properly with QFutures


LightZam:
I have a problem about sending an iq.

for example,

```
QXmppMucRoom *room = client->mucManager().rooms().value(0);
room->setConfiguration(form);
```

there is no possible way to know setConfiguration(form)'s result, 

because iq's id was hidden from user who use the lib in this situation, and QxmppMucRoom didn't catch the result for user. so there is no way to know setConfiguration works or not.

is it a design issue in this case?

how to get result from iq if i need to know?

jlaine:
The API could indeed be improved here. However, having the user deal directly with sending / receiving the IQs kind of defeats the whole "manager" concept. Any suggestions?

LightZam:
how about using lambda or promise to handle sendPacket ?
it might help a lot when deal with receiving every iq's result.

lnjX:
I thought of something like this (not sure whether everything)
```cpp
class QXmppClient {
    QSharedPointer<QXmppIqResult> sendIq(QXmppIq iq);
}
class QXmppClientPrivate {
    QVector<QSharedPointer<QXmppIqResult>> pendingIqResults;
}

class QXmppIqResult {
    enum TransmissionError {
        Disconnected, // disconnectFromServer() called, before result could arrive
        SocketFailure // IQ was sent although socket was not connected or did not send the full data (see  https://github.com/qxmpp-project/qxmpp/blob/master/src/base/QXmppStream.cpp#L132-L138 )
    };

signals:
    void replyReceived(QDomElement element);
    void errorReceived(QXmppIq iq);
    void transmissionFailed(TransmissionError error)

public:
    void resendIq(); // resend same request, but with a new ID (not sure if that is really needed)
}
```

It would also be nice to get a feedback, when a `<message/>` was really sent / acknowledged via stream management.

0xd34df00d:
I'd vote for `QFuture`-based API. It just composes a bit better with other primitives that might be out there for working with and chaining futures (or at least my codebase does have those :)).

lnjX:
How would I create a QFuture? Obviously `QtConcurrent::run()` is not what we want. I found something about `QFutureInterface`, but I can't find any documentation for it.

0xd34df00d:
[...]

PR #102, opened 2016-08-14
Allow to receive IQ messages other than for the configuration

kollix:
To fix issue #96 this small change shall be included, please.
To know the QXmppIq.id() for the result message, I create my own QXmppMucOwnerIq
instead of using room->setConfiguration(form), e.g.:

  QXmppMucOwnerIq iqPacket;
  iqPacket.setType(QXmppIq::Set);
  iqPacket.setTo(room->jid());
  iqPacket.setForm(form);
  resultId = iqPacket.id();

then I wait for the IQ message:

```
connect(client, SIGNAL(iqReceived(const QXmppIq &)),
        this, SLOT(iqReceived(const QXmppIq &)));
```

and in the slot:

void XmppWaitMucRoom::iqReceived(const QXmppIq &iq)
{
  if ( (iq.from() != room->jid()) ||
       (iq.id() != resultId) )
    return;

  if ( iq.type() == QXmppIq::Error )
    error(iq.error());

  done = true;
}


lnjX:
Replaced by #226.

PR #103, opened 2016-08-14
allow to compile with a namespaced installation of Qt

kollix:
This patch allows to compile the code with a Qt installation, which has been compiled with a namespace (Qt configure called with -qtnamespace XXX -qtlibinfix XXX),
which is needed when an application itself delivers the qt libs so that they do not conflict with the system installed ones (as our commercial product does).

By using the Qt macro QT_FORWARD_DECLARE_CLASS one can compile the same code with or without Qt being namespaced.

### DIFF
--- a/examples/example_3_transferHandling/example_3_transferHandling.h
+++ b/examples/example_3_transferHandling/example_3_transferHandling.h
@@ -28,8 +28,6 @@
 #include "QXmppClient.h"
 #include "QXmppTransferManager.h"
 
-class QBuffer;
-
 class xmppClient : public QXmppClient
 {
     Q_OBJECT
--- a/src/base/QXmppElement.h
+++ b/src/base/QXmppElement.h
@@ -30,7 +30,7 @@
 
 #include "QXmppGlobal.h"
 
-class QDomElement;
+QT_FORWARD_DECLARE_CLASS(QDomElement)
 class QXmppElement;
 class QXmppElementPrivate;
 
--- a/src/base/QXmppSocks.h
+++ b/src/base/QXmppSocks.h
@@ -29,7 +29,7 @@
 
 #include "QXmppGlobal.h"
 
-class QTcpServer;
+QT_FORWARD_DECLARE_CLASS(QTcpServer)
 
 class QXMPP_EXPORT QXmppSocksClient : public QTcpSocket
 {
--- a/src/base/QXmppStream.h
+++ b/src/base/QXmppStream.h
@@ -30,8 +30,8 @@
 #include <QObject>
 #include "QXmppLogger.h"
 
-class QDomElement;
-class QSslSocket;
+QT_FORWARD_DECLARE_CLASS(QDomElement)
+QT_FORWARD_DECLARE_CLASS(QSslSocket)
 class QXmppStanza;
 class QXmppStreamPrivate;
 
--- a/src/base/QXmppStun.h
+++ b/src/base/QXmppStun.h
@@ -30,9 +30,9 @@
 #include "QXmppJingleIq.h"
 
 class CandidatePair;
-class QDataStream;
-class QUdpSocket;
-class QTimer;
+QT_FORWARD_DECLARE_CLASS(QDataStream)
+QT_FORWARD_DECLARE_CLASS(QUdpSocket)
+QT_FORWARD_DECLARE_CLASS(QTimer)
 class QXmppIceComponentPrivate;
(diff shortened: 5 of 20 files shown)


examples/example_3_transferHandling/example_3_transferHandling.h
```diff
@@ -28,7 +28,7 @@
 #include "QXmppClient.h"
 #include "QXmppTransferManager.h"
 
-class QBuffer;
+#include <QBuffer>
```
jlaine: Why don't we use QT_FORWARD_DECLARE_CLASS here?

jlaine:
Whilst I have no problem with supporting a namespaced Qt per se, it's the kind of thing that is guaranteed to break unless it's systematically testing by CI. Any suggestions on how to do this?

Also : using travis with Ubuntu 12.04 had gone to hell completely, I switched to the 14.04 image, which does not require any additional PPAs. Could you please rebase on top of my latest commit?

examples/example_3_transferHandling/example_3_transferHandling.h
kollix: On Monday 15 August 2016 15:29:28 Jeremy Lain� wrote:
> > @@ -28,7 +28,7 @@
> >  #include "QXmppClient.h"
> >  #include "QXmppTransferManager.h"
> >  
> > -class QBuffer;
> > +#include <QBuffer>
> 
> Why don't we use QT_FORWARD_DECLARE_CLASS here?

There are 2 ways to fix the issue with namespaced Qt:
1) use QT_FORWARD_DECLARE_CLASS when doing forward declares
2) include the class, since the Qt headers deal with the namespace (with the drawback of including a file)

I think in that case I ran my "find/sed" call only across the src directory but not across examples
and when compiling everything and hit the error, I just modified this example by hand, and since
it's just an example, it does not matter if I use approach 1 or 2

But in fact I found now that the QBuffer in this example header file
examples/example_3_transferHandling/example_3_transferHandling.h
is not needed at all.
So the simplest fix is: remove the "class QBuffer;" entirely

-- 
Best regards/Sch�ne Gr��e

Martin
Q: Why is top posting bad?

kollix:
On Monday 15 August 2016 15:31:57 Jeremy Lain� wrote:
> Whilst I have no problem with supporting a namespaced Qt per se, it's the kind of thing that is guaranteed to break unless it's systematically testing by CI. Any suggestions on how to do this?

Without a namespaced Qt installation, you can not entirely automate a test.
But in fact I see it much simpler:
How many bug reports did you get for QXmpp not working with a namespaced Qt ?
I assume: 0
Which simply means, I'm the only one hit by this problem, which means: whenever in the future
I will recompile a new version locally, I will produce another patch if something was broken.

> Also : using travis with Ubuntu 12.04 had gone to hell completely, I switched to the 14.04 image, which does not require any additional PPAs. Could you please rebase on top of my latest commit?

I'm new to git ... hope I did that now correctly.

-- 
Best regards/Sch�ne Gr��e

Martin
Q: Why is top posting bad?

Kaffeine:
@kollix :
```
git checkout namespace # Ensure that you're on the branch
git reset --hard 8606658a9c9eef83e89e6f62fffd913bffb99b0f # Reset the branch to latest commit from origin.
# If you have this (qxmpp-project) repo in git configuration, then you can reset to the origin master as well (git reset --hard origin/master ).
git cherry-pick e99dea733704e6f5f54a5db4069c48acc79c8ca0 # Re-apply your fix
git cherry-pick 24c91c5e6d31dc9ead27fe773daab3eea2e8e46f # Re-apply your fix to the fix :)
git rebase -i HEAD^^  # Ask git to edit commits.
```
Now change the "pick" word in the second line to "f" to fixup the last commit to the previous one.

was been (commits identifiers would be different):
```
pick eb65e6b allow to compile with a namespaced installation of Qt
pick 1d265fa remove unused include
```

Should be:
```
pick eb65e6b allow to compile with a namespaced installation of Qt
f 1d265fa remove unused include
```

Change, save and exit.
You can check your changes via ```git log -p```

It's time to update the PR. I have no idea how you setup git, so just do this:
```
git push -f https://github.com/kollix/qxmpp.git HEAD:namespace
```

Depends on configuration, it can be enough to just do ```git push -f```, but the first option is 100% right.

After this commands, you can check this PR. It should be updated automatically.

kollix:
ok, did that now

lnjX:
If there is still interest in this, please reopen and rebase.

PR #104, opened 2016-08-17
In Stanza error make to/from string methods public

kollix:
I'm integrating XMPP into a scripting language. In this context it is useful to have the stanza error type and condition as strings, therefore this patch makes the existing methods public.

### DIFF
--- a/src/base/QXmppStanza.h
+++ b/src/base/QXmppStanza.h
@@ -149,13 +149,13 @@ public:
         void toXml(QXmlStreamWriter *writer) const;
         /// \endcond
 
-    private:
         QString getConditionStr() const;
         void setConditionFromStr(const QString& cond);
 
         QString getTypeStr() const;
         void setTypeFromStr(const QString& type);
 
+    private:
         int m_code;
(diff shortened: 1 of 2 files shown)


jlaine:
I am not in favor of this: we have canonical representations for this data in the form of enums, I do not want an alternative representation made public.

kollix:
On Wednesday 17 August 2016 03:32:55 Jeremy Lain� wrote:
> I am not in favor of this: we have canonical representations for this data in the form of enums, I do not want an alternative representation made public.

ok, I understand. Then I'll make some internal copy of them

-- 
Best regards/Sch�ne Gr��e

Martin
Q: Why is top posting bad?

Kaffeine:
This issue/pull request can be closed then.

For anyone who would like to copy the toString method I would suggest to copy it from this [project](https://github.com/TelepathyQt/telepathy-nonsense/blob/72366430ad7afd3cb1481952eb65ded4afa4e160/textchannel.cc#L24) with a bit better (for human reading) text.

PR #105, opened 2016-08-17
automatically try next server from DNS SRV record when connection to first can not be established

kollix:
I had the problem (company network) where the default port 5222 is blocked.
The DNS record however knows an alternative server, which was not used automatically.

This patch adds the possibility to automatically try the next server when the first fails.

(I hope the check d->sessionStarted is correct)

### DIFF
--- a/src/client/QXmppOutgoingClient.cpp
+++ b/src/client/QXmppOutgoingClient.cpp
@@ -64,6 +64,7 @@ class QXmppOutgoingClientPrivate
 public:
     QXmppOutgoingClientPrivate(QXmppOutgoingClient *q);
     void connectToHost(const QString &host, quint16 port);
+    void connectToNextDNSHost();
 
     void sendNonSASLAuth(bool plaintext);
     void sendNonSASLAuthQuery();
@@ -78,6 +79,7 @@ public:
 
     // DNS
     QDnsLookup dns;
+    int nextSrvRecordIdx;
 
     // Stream
     QString streamId;
@@ -117,7 +119,8 @@ private:
 };
 
 QXmppOutgoingClientPrivate::QXmppOutgoingClientPrivate(QXmppOutgoingClient *qq)
-    : redirectPort(0)
+    : nextSrvRecordIdx(0)
+    , redirectPort(0)
     , bindModeAvailable(false)
     , sessionAvailable(false)
     , sessionStarted(false)
@@ -160,6 +163,15 @@ void QXmppOutgoingClientPrivate::connectToHost(const QString &host, quint16 port
     } else {
         q->socket()->connectToHost(host, port);
     }
+}
+
+void QXmppOutgoingClientPrivate::connectToNextDNSHost()
+{
+    connectToHost(
+        dns.serviceRecords().at(nextSrvRecordIdx).target(),
+        dns.serviceRecords().at(nextSrvRecordIdx).port());
+
+    nextSrvRecordIdx++;
 }
 
 /// Constructs an outgoing client stream.
@@ -251,6 +263,7 @@ void QXmppOutgoingClient::connectToHost()
     d->dns.setName("_xmpp-client._tcp." + domain);
     d->dns.setType(QDnsLookup::SRV);
     d->dns.lookup();
+    d->nextSrvRecordIdx = 0;
 }
 
 void QXmppOutgoingClient::disconnectFromHost()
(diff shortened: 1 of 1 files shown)


Kaffeine:
Hello @kollix ,
I would like to try to make it merged, but we need to polish it first.

IMO the first step would be to fixup 24c91c5e6d31dc9ead27fe773daab3eea2e8e46f into e99dea733704e6f5f54a5db4069c48acc79c8ca0 and then we can pick this commit as "Fixed compilation with a namespaced Qt".

Second thing: commit 087722f12895ddb85abe0beccfc313977cf476c6. The commit adds internal debug methods to public API/ABI. I found them useful, but decided to write own method with a bit more readable texts: https://github.com/TelepathyQt/telepathy-nonsense/commit/72366430ad7afd3cb1481952eb65ded4afa4e160 Developers can be against this commit, so it would be better to make another PR, but let's wait for other opinions.

After that, we should rebase this PR to be just two commits:
02e2fa3b0729dacbfff0e5354c1ef94477ea3f7d (Allow to receive IQ messages other than for the configuration) and 66ba2fea0569661b9567282814d74415e082c415 (Try next server from DNS record when last one fails to start session).
I think that it would be better to extract 
```cpp
       d->connectToHost(
            d->dns.serviceRecords().at(d->nextSrvRecordIdx).target(),
            d->dns.serviceRecords().at(d->nextSrvRecordIdx).port());
        d->nextSrvRecordIdx++;
```
and use it in both _q_dnsLookupFinished() and socketError() to reduce code duplication.

kollix:
I obviously messed up the git branch.
Hope this commit is now good enough for a merge

src/client/QXmppOutgoingClient.cpp
```diff
@@ -287,7 +298,14 @@ void QXmppOutgoingClient::socketSslErrors(const QList<QSslError> &errors)
...
     Q_UNUSED(socketError);
-    emit error(QXmppClient::SocketError);
+    if ( !d->sessionStarted &&
+         (d->dns.serviceRecords().count() > d->nextSrvRecordIdx) )
+    {
+        // some network error occured during startup -> try next available SRV record server
+        d->connectToNextDNSHost();
+    }
+    else
+      emit error(QXmppClient::SocketError);
```
Kaffeine: Nitpick: ```emit error``` has wrong indentation.

Kaffeine:
Other than the nitpick, this commit looking good to me now.
(I'm not the project developer)

Kaffeine:
@kollix You didn't say it and I forgot to ask: does this patch fix the problem? It would be good to know that you tested it :)

kollix:
On Sunday 04 September 2016 11:44:13 Alexandr Akulich wrote:
> Other than the nitpick, this commit looking good to me now.
> (I'm not the project developer)

I fixed the indent now

-- 
Best regards/Sch�ne Gr��e

Martin
Q: Why is top posting bad?

kollix:
On Sunday 04 September 2016 11:48:35 Alexandr Akulich wrote:
> @kollix You didn't say it and I forgot to ask: does this patch fix the problem? It would be good to know that you tested it :)
> 
> 

yes, I tested it and it works.

-- 
Best regards/Sch�ne Gr��e

Martin
Q: Why is top posting bad?

Kaffeine:
@kollix Fix is correct, but now you have two commits instead of one.

After made the fix, you could do
```
git add -u # Add the change
git commit --amend --no-edit # "Amend" means that you want to change the last commit, "no edit" means that you want to keep commit message the same.
git push -f # Push the changes to the PR branch on github and overwrite it with the local version
```

Now you can either do
```
git rebase -i HEAD^^ # Rebase the last two commits
# Change the first word in the second line to "f" or "fixup" to merge the last change (indentation fix) into the main change ("automatically try next server....").
git push -f # Push the changes to the PR branch on github and overwrite it with the local version
```
or ```git reset HEAD^``` and then the first commands, start with ```git add -u```.

kollix:
On Sunday 04 September 2016 12:16:15 Alexandr Akulich wrote:
> @kollix Fix is correct, but now you have two commits instead of one.

yes. And why is this bad ?
It's a bit cumbersome to always change the git history just to have 1 commit.

-- 
Best regards/Sch�ne Gr��e

Martin
Q: Why is top posting bad?

Kaffeine:
Following is my personal opinion, which can not fit any use case and project.

>> now you have two commits instead of one.

>yes. And why is this bad ?

Because if want someone to review and accept your change, then it is good to clear changes to easy the process.
There is a number of good practice, e.g. it is good to not mix code format changes with actual code change.

It is not a problem to have a series of commits, but it makes no sense to write in a history of project that you made some spacing fix to your own previous commit.

I would suggest to look at commits of [Linux](https://lkml.org/), [Mesa](https://lists.freedesktop.org/archives/mesa-dev/), [Qt](https://codereview.qt-project.org/) or any other big and high quality project for the best commit practice. There is no known 100% perfect way, but IMO those project are good enough to be a reference.

PR #115, opened 2016-10-09
QT_NO_KEYWORDS patch

tomcucinotta:
Dear QXMPP-ers,

I'm proposing a patch to switch to a QT_NO_KEYWORDS paradigm when using Qt in QXMPP. I had to use such a modified version in order to use QXMPP for enhancing LyX with chat capabilities:
http://www.lyx.org/trac/ticket/7964

Unfortunately, using the regular QXMPP header files clashes with boost libraries, whilst switching to the Q_\* variant (which is a Qt supported syntax anyway) makes it possible & smooth to use QXMPP along with boost libraries. This seems a good thing, in terms of usability of your library within other projects. Please, let me know what you think!

Thanks & keep up with the excellent work :-)!

### DIFF
--- a/examples/GuiClient/capabilitiesCache.h
+++ b/examples/GuiClient/capabilitiesCache.h
@@ -47,9 +47,9 @@ public:
     QStringList getFeatures(const QString& nodeVer);
     QStringList getIdentities(const QString& nodeVer);
 
-signals:
+Q_SIGNALS:
 
-private slots:
+private Q_SLOTS:
     void infoReceived(const QXmppDiscoveryIq&);
 
 private:
--- a/examples/GuiClient/chatDialog.h
+++ b/examples/GuiClient/chatDialog.h
@@ -53,7 +53,7 @@ public:
     void setQXmppClient(QXmppClient* client);
     void messageReceived(const QString& msg);
 
-private slots:
+private Q_SLOTS:
     void sendMessage();
 
 protected:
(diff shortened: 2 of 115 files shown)


jlaine:
I can possibly understand the rationale for the headers, but why change the code itself?

tomcucinotta:
Hi,

I just imagined one wouldn't like to have a mix of headers with a Q_SLOTS with implementation code with slots:, and I don't really know whether they are really compatible / mixable in Qt.

FYI, the patch has been obtained by just running smth like:

find . \( -name '*.h' -o -name '*.cpp' \) -exec sed -i -e 's/slots:Q_SLOTS:/' -e 's/signals:/Q_SIGNALS' -e 's/emit /Q_EMIT /' {} \;

Thanks!

jlaine:
I'm sorry, but with our limited resources I don't see us merging this

tehnick:
`Q_SLOTS` looks ugly in comparing with `slots`. The same is for other replacements...

As for `emit` -- it does nothing at all now. Just a pretty syntactic sugar...

tomcucinotta:
On 21/01/19 10:48, Boris Pek wrote:
>
> |Q_SLOTS| looks ugly in comparing with |slots|. The same is for other replacements...
>
> As for |emit| -- it does nothing at all now. Just a pretty syntactic sugar...
>
sure, but please understand the rationale behind my original post: this is not about keeping the code nice & cute, this 
is just about being able to compile this software along with others, in a Qt application where QT_NO_KEYWORDS was 
assumed in other headers. I'll have to double-check what is the situation now, as some time (>1y?) elapsed since last 
time I compiled this...

Thanks.

     T.

-- 
Tommaso Cucinotta
Home Page: http://retis.sssup.it/~tommaso
LinkedIn: http://www.linkedin.com/in/tommasocucinotta

lnjX:
Replaced by #258.

PR #121, opened 2017-02-02
Tiny cosmetic fix to avoid cppcheck message about memory leak

tehnick:


tehnick:
PR is updated.

0xd34df00d:
While this change indeed improves readability of the code, I'd argue that cppcheck is probably not the best tool out there for static analysis, and in general code shouldn't be modified just for the sake of fixing its warnings.

0xd34df00d:
Also, a better fix would be to use std::unique_ptr in C++11 world, or std::vector in 03.

PR #122, opened 2017-02-02
Fix cppcheck warnings

tehnick:


examples/GuiClient/chatGraphicsScene.cpp
```diff
@@ -27,7 +27,7 @@
 #include "chatGraphicsView.h"
 
 chatGraphicsScene::chatGraphicsScene(QObject* parent) : QGraphicsScene(parent),
-        m_verticalPosForNewMessage(0), m_verticalSpacing(5)
+        m_verticalPosForNewMessage(0), m_verticalSpacing(5), m_boxStartLength(5)
```
0xd34df00d: Where does this value come from?

src/base/QXmppJingleIq.cpp
```diff
@@ -710,7 +710,7 @@ class QXmppJingleIqPrivate : public QSharedData
 };
 
 QXmppJingleIqPrivate::QXmppJingleIqPrivate()
-    : ringing(false)
+    : action(QXmppJingleIq::ContentAccept), ringing(false)
```
0xd34df00d: ditto

src/base/QXmppRtpChannel.h
```diff
@@ -209,6 +209,9 @@ class QXMPP_EXPORT QXmppVideoFrame
 class QXMPP_EXPORT QXmppVideoFormat
 {
 public:
+    QXmppVideoFormat() : m_frameRate(15.0), m_frameSize(QSize(320, 240)),
+        m_pixelFormat(QXmppVideoFrame::Format_YUYV) {}
```
0xd34df00d: ditto

src/client/QXmppCallManager.cpp
```diff
@@ -104,7 +104,8 @@ class QXmppCallManagerPrivate
 };
 
 QXmppCallPrivate::QXmppCallPrivate(QXmppCall *qq)
-    : manager(0),
+    : direction(QXmppCall::IncomingDirection),
```
0xd34df00d: ditto

src/base/QXmppElement.cpp
```diff
@@ -123,10 +123,13 @@ QXmppElement::~QXmppElement()
 
 QXmppElement &QXmppElement::operator=(const QXmppElement &other)
 {
-    other.d->counter.ref();
-    if (!d->counter.deref())
-        delete d;
-    d = other.d;
+    if (this != &other) // self-assignment check
```
0xd34df00d: This check doesn't actually affect correctness, right?

If so, I'm not sure it should be there. The performance improvements might be marginal, if any. Also, note that, say, typical std::string implementations don't do any self-assignment checks in their operator= (at least, I'm not aware of any implementations doing so).

src/base/QXmppRtpChannel.h
tehnick: https://github.com/qxmpp-project/qxmpp/blob/034c318/src/base/QXmppRtpChannel.cpp#L809

src/client/QXmppTransferManager.cpp
```diff
@@ -220,6 +220,7 @@ class QXmppTransferJobPrivate
 QXmppTransferJobPrivate::QXmppTransferJobPrivate()
     : blockSize(16384),
     client(0),
+    direction(QXmppTransferJob::IncomingDirection),
```
tehnick: First element in enum.

src/client/QXmppCallManager.cpp
tehnick: First element in enum.

src/base/QXmppJingleIq.cpp
tehnick: First element in enum.

examples/GuiClient/chatGraphicsScene.cpp
tehnick: From nowhere. The main idea was that it should not be equal to null.

src/base/QXmppElement.cpp
tehnick: > This check doesn't actually affect correctness, right?

Yep.

> The performance improvements might be marginal

Are you sure? This looks quite long:
https://github.com/qxmpp-project/qxmpp/blob/034c318/src/base/QXmppElement.cpp#L53

tehnick: > This check doesn't actually affect correctness, right?

Yep.

> The performance improvements might be marginal, if any.

I agree.

> If so, I'm not sure it should be there. 

So am I. But I also do not see reasons why not.

src/base/QXmppRtpChannel.h
jlaine: On a stylistic point, let's please align with Qt:

class Foo
    : foo(0)
    , bar(1)

src/base/QXmppElement.cpp
jlaine: Qt does no self assignment checks either

tehnick: 1) Qt uses smart pointers (QSharedDataPointer, etc) in many classes.
2) There are few places in Qt there self assignment checks are present. For example:
http://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/tools/qcollator.cpp?h=v5.8.0#n107
http://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/io/qdebug.h?h=v5.8.0#n200
http://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/kernel/qvariant.cpp?h=v5.8.0#n1881
http://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/text/qtextoption.cpp?h=v5.8.0#n122
etc.

tehnick:
> I am a little concerned by vast sweeping changes which don't seem to correct any actual bugs.

Do you think it is better to leave uninitialized variables in the *library*? Even if they will not cause segfaults due to specifics of usage of these classes, it is bad idea IMHO.

tehnick:
> On a stylistic point, let's please align with Qt:

Done.

src/base/QXmppElement.cpp
tehnick: > Also, note that, say, typical std::string implementations don't do any self-assignment checks in their operator= (at least, I'm not aware of any implementations doing so).

Why have you mentioned `std::string`? I do not know about it, but other STL classes do have self-assignment checks in `operator=`. At least in libstdc++.

tehnick: And of course I may remove this commit from PR if you are against it.

tehnick:
Rebased against master.

Any additional comments?

tehnick:
Any updates on this thread?

tehnick:
Please merge or reject.

0xd34df00d:
@jlaine while I agree vast changes might introduce bugs, fixing even false positives from tools like cppcheck (even though they are IMO crappy compared to clang-tools/coverity) reduces the noise on subsequent runs and allows spotting the actual detected bugs more easily.

So, I'd like to merge this if you don't mind.

jlaine:
If you could just add a test case for the self assignment you're fixing, go ahead and merge Boris.

tehnick:
PR is rebased against current master.

tehnick:
cppcheck logs for current master:
https://tehnick.net/sibuserv/projects/qxmpp/v0.9.3-49-g09560c9/StaticCodeAnalysis/cppcheck.log
https://tehnick.net/sibuserv/projects/qxmpp/v0.9.3-49-g09560c9/StaticCodeAnalysis/cppcheck.html/

Issue #95, opened 2016-02-29
Capital letters in jid


fbeutel:
I recently came across an error when using QXmppMucManager:

When adding a room with a JID that contains capital letters (e.g. `mucmanager->addRoom("TeSt@example.com")`, QXmpp doesn't recognize a successful join, because the server response presence is being sent using lower letters ("test@example.com"), so the jid comparisons don't work.

Since jid comparisons are probably not very uncommon in QXmpp, this might be hiding at other places, too.

Now I'm not sure what the best solution would be. Obviously, QXmpp could demand of the users that jids be always using lower letters, but that doesn't seem right for several reasons.
First, it is still prone to errors like the one above if the user isn't super careful - it's too easy to introduce such bugs.
Second, couldn't a server implementation (in theory) also use upper letters in it's responses?

Instead, maybe the easiest way is to add a jid comparison function to QXmppUtils and consistantly use that for comparisons.
Another option would be to introduce some kind of case-insenitive QString-compatible class to store JIDs and everywhere, where we store a JID, use that class instead of QString...

What do you think?

0xd34df00d:
I had the very same issue a couple of weeks ago. I fixed it by just
lowercasing all the room JIDs in the right place of my client's code.

I could not find anything in XEP-0045 or anywhere else about the
requirement of lowercasing the JIDs, though, so I'm not sure it's the right
thing to implement in QXmpp itself.

Anyway, having a separate JID class seems like a good solution, bringing
more type safety.

Note that not only comparisons will have to be case-insensitive, but also
the hash function as well.

jlaine:
On 02/29/2016 09:14 PM, Georg Rudoy wrote:
> I had the very same issue a couple of weeks ago. I fixed it by just
> lowercasing all the room JIDs in the right place of my client's code.
>
> I could not find anything in XEP-0045 or anywhere else about the
> requirement of lowercasing the JIDs, though, so I'm not sure it's the
> right
> thing to implement in QXmpp itself.
>

I had a look at RFC 3920, and it would seem:

- the "node" (~username) part of the JID is indeed case insensitive. The
NODEPREP profile refers to STRINGPREP tables B.1 and B.2 for mapping

- the "resource" part of the JID is case sensitive. The RESOURCEPREP
profile refers only to STRINGPREP table B.1 for mapping.

> Anyway, having a separate JID class seems like a good solution, bringing
> more type safety.
>

I am not super fond of introducing a JID class, as it's too late to use
[...]

PR #126, opened 2017-04-21
Fix : MUC Room joined signal can not be emitted

AdityaKurude:
All names in qmpp are received in lower cases and hence MUC room names should also be converted to lower case while they are being created. If user creates a room having any capital letter and requests to join the room, joined signal is not emitted because of the mismatch between cases, This code fixes the bug where joined() signal is not emitted even after room->join API is invoked.

### DIFF
--- a/src/client/QXmppMucManager.cpp
+++ b/src/client/QXmppMucManager.cpp
@@ -75,10 +75,10 @@ QXmppMucManager::~QXmppMucManager()
 
 QXmppMucRoom *QXmppMucManager::addRoom(const QString &roomJid)
 {
-    QXmppMucRoom *room = d->rooms.value(roomJid);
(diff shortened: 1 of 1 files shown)


jlaine:
I am not sure I understand the root cause, who is causing the room to be lowercased?

AdityaKurude:
While testing with Prosody server, I came across this use case wherein when I create a room having a name with Uppercase letter, when room join is requested there is mismatch between the locally stored roomName in QXmppMucRoomPrivate and jid received from the server. Here, jid received from the server is always lower cased value.

lnjX:
I don't think that's the correct way to solve this, if we needed to change this (e.g. force everything to be lower cased) this would need to be done for all JIDs, everywhere in QXmpp and not only for MUC joining. See https://github.com/qxmpp-project/qxmpp/issues/95.

PR #131, opened 2017-09-04
Add CMake build system, drop Qt4 support

olesalscheider:
This patch series adds a CMake based build system. It also drops the Qt4 support.

Once this is merged, I want to clean up and modernize the code base.
This includes things like the new Qt5 connect syntax, QLatin1String/QStringLiteral and C++11 features.
I also want to get a coherent coding style.


jlaine:
Hi Niels,

Why a pull request for two unrelated goals?

Dropping Qt4: it's probably time indeed. It's worth sending a message to the mailing list though, see if there are any strong reasons not to do it.

Using Cmake as the build system: why is this a goal at all? I'm fond of Cmake but unfortunately it's just not as ubiquitous as qmake among Qt users (hi windows, android, iOS..). I'd be open to adding support for *using* QXmpp in a Cmake project but I honestly don't see the value in changing the internal build system.

0xd34df00d:
I second @jlaine. One more point, though: what's your opinion on more modern C++11 features? Now that recent Qt 5 releases require C++11-enabled compilers, especially.

jlaine:
Concerning GuiClient it's the only decent fully featured example we have. If memory serves it was cut out of the build because it was not ported to Qt5. Maybe a separate repo ?

jlaine:
Concerning C++11 I'd say once we make QXmpp Qt5 only adopt the same policy as Qt. I believe it's something like: "use any of the C++11 features supported by all compilers internally, but don't let it leak at the API boundary"

olesalscheider:
> Why a pull request for two unrelated goals?

Only because I did not want to touch the qmake build system just to drop Qt4 support. It was really was dropped by not supporting it with the CMake based and then I just removed the remains.
But I can split it, if you want.

> Dropping Qt4: it's probably time indeed. It's worth sending a message to the mailing list though, see if there are any strong reasons not to do it.

Ok, I will do that. I do not really see a problem here. Platforms that are stuck at Qt4 probably won't have a problem with being stuck at the current QXmpp release (to be honest there is not that much development going on nowadays...)

> Concerning GuiClient it's the only decent fully featured example we have. If memory serves it was cut out of the build because it was not ported to Qt5. Maybe a separate repo ?

I can re-add it, if you want. But a separate repo sounds like a good idea anyway.

> Concerning C++11 I'd say once we make QXmpp Qt5 only adopt the same policy as Qt. I believe it's something like: "use any of the C++11 features supported by all compilers internally, but don't let it leak at the API boundary"

I agree. My plan would be to follow this policy. There is no need to use some exotic constructs, but there are some prominent things that make life much easier.

olesalscheider:
> Using Cmake as the build system: why is this a goal at all? I'm fond of Cmake but unfortunately it's just not as ubiquitous as qmake among Qt users (hi windows, android, iOS..). I'd be open to adding support for using QXmpp in a Cmake project but I honestly don't see the value in changing the internal build system.

IMO CMake is just a much nicer/cleaner build system. I think I do not have to list all the general advantages here...

And with this patchset we get real improvements. See for example how we now configure QXmppGlobal.h so that we only have to set the version in the main CMakeLists.txt. Or how we get pkg-config support on Linux for the multimedia libraries. Or how it makes lifes much easier for distributions with a more complex file system layout (cf. https://github.com/qxmpp-project/qxmpp/pull/49).

Also I'm not sure that qmake is more popular along the Qt users (cf. the poll in [1]). It will also be deprecated (but still supported) with Qt6 which is coming closer. Qt6 will probably either move to Qbs or CMake.

Apart from that, it is really easy to build CMake based projects on Windows. Yes, you have to download the executable and install it, but I think that is something that we can expect from a developer? And once you have done this it only makes your life easier: For example, if it can't determine the path to some dependency (like the multimedia libs) you can just use the CMake GUI to point it to the right location...

And you can also build Android and iOS applications with CMake.
I don't have any experience with that myself but you can just add corresponding targets to the build system that do everything for you. There are other projects that do the same.

[1] https://wiki.qt.io/Qt_build_systems_at_QtCon_2016

jlaine:
I am convinced that cmake is a superior build system to qmake - I'm definitely not disputing this, and I used to maintain a CMake-based build of QXmpp. However, as said it's not ubuitous as in, it requires an additional installation, and specific steps to build QXmpp, which raises the entry bar for new users. At the moment the *only* thing needed to build QXmpp is a Qt SDK.

Having two build systems in parallel is a non-starter, it increases the maintenance burden and you can be sure that the qmake build will fall in disrepair.

The only way I can see us switching to CMake is if we provide (and test!) binary releases of QXmpp for all non-Linux platforms supported by Qt. This would be nice, although I'm not exactly sure whether travis (or some alternative free service) provides the infrastructure to do so. I'm less concerned about the Linux as:

a/ some distros ship ready-to-use builds of QXmpp
b/ they tend to be more familiar with tooling

staniek:
Hi, At least [FindQxmpp](https://github.com/0xd34df00d/leechcraft/blob/master/src/plugins/azoth/plugins/xoox/cmake/FindQXmpp.cmake) would be good to have in this project, otherwise cmake-based projects are unable to find the lib as easily as they can find Qt or other libs i.e. in one line of code. In the link above you can see unofficial files like that exist.

olesalscheider:
Hi Jeremy,

thanks... I do not have push rights though.

I think I can set up some automated builds or at least add scripts for easy builds on other platforms - I am working on something that is supposed to also run on android and I will have to do this anyway.

jlaine:
Sorry about that I was convinced I had given you the rights. You should have received an invitation now.

olesalscheider:
Thanks :)

PR #134, opened 2017-11-30
get resume id from server

coderlxn:


olesalscheider:
LGTM.

jlaine:
IQ parsing / serialization is trivial to unit test, so do it :)

jlaine:
@coderlxn can we please have a unit test for this so we can merge it?

jlaine:
@lnjX I'm not sure I understand what this does anymore, ideas?

src/base/QXmppStreamManagement.cpp
```diff
@@ -132,6 +132,7 @@ void QXmppStreamManagementEnabled::parse(const QDomElement &element)
     m_resume = resume == QString("true") || resume == QString("1");
     m_max = element.attribute("max").toUInt();
     m_location = element.attribute("location");
+    m_id = element.attribute("id");
```
lnjX: This only adds parsing, we also need to add this in `::toXml`.

lnjX:
@jlaine The session id needs to be parsed to be able to resume the session (using `previd` in `<resume/>`) after a possible connection loss.

lnjX:
Unfortunately there's no test class for stream management, yet, but you could base on [this test](https://github.com/qxmpp-project/qxmpp/blob/master/tests/qxmppiq/tst_qxmppiq.cpp) basically. It were also nice to have this new test class in a new commit.

lnjX:
Parsing of the id attribute has been added in the meantime, so this is obsolete.

PR #139, opened 2018-05-15
Add support for NV21 video format in vpx encoder

hhaveri:


jlaine:
Cool! I'm starting to wonder whether we should break out the format conversion code so that we can add unit tests. Thoughts?

hhaveri:
No thoughts at the moment if you're asking from me :)

jlaine:
This PR contains two unrelated changes, on of which is a breaking API change. Can you please only include the NV12 support in this PR, break it out to a separate function and provide a unit test for it as requested?

lnjX:
I'm closing this due to inactivity and because we probably don't need this anymore as soon as #207 is done.

PR #142, opened 2018-08-13
support XEP-0237 Roster Versioning

LightZam:
# Description
add version attribute to support protocol XEP-0237 Roster Versioning

# Protocol Reference
https://xmpp.org/rfcs/rfc6121.html#roster-versioning
https://xmpp.org/extensions/xep-0237.html


LightZam:
@jlaine is this feature looks good to you?

Issue #141, opened 2018-06-29
tst_qxmppversioniq.moc: No such file or directory


vasilyrud:
When compiling on Ubuntu 16.04 I get the following error:

```
[ 34%] Building CXX object tests/CMakeFiles/tst_qxmppversioniq.dir/qxmppversioniq/tst_qxmppversioniq.cpp.o
/home/vasily/xmpp-fuzzer/qxmpp/tests/qxmppversioniq/tst_qxmppversioniq.cpp:79:34: fatal error: tst_qxmppversioniq.moc: No such file or directory
compilation terminated.
tests/CMakeFiles/tst_qxmppversioniq.dir/build.make:62: recipe for target 'tests/CMakeFiles/tst_qxmppversioniq.dir/qxmppversioniq/tst_qxmppversioniq.cpp.o' failed
make[2]: *** [tests/CMakeFiles/tst_qxmppversioniq.dir/qxmppversioniq/tst_qxmppversioniq.cpp.o] Error 1
CMakeFiles/Makefile2:273: recipe for target 'tests/CMakeFiles/tst_qxmppversioniq.dir/all' failed
make[1]: *** [tests/CMakeFiles/tst_qxmppversioniq.dir/all] Error 2
Makefile:138: recipe for target 'all' failed
make: *** [all] Error 2
```

Steps to reproduce:

```
git clone https://github.com/qxmpp-project/qxmpp.git
cd qxmpp
mkdir build
cd build
cmake ..
make
```

Some details:

`cmake version 3.5.1`
`GNU Make 4.1`
`Linux version 4.4.0-116-generic`
`Using Qt version 5.5.1`

LightZam:
if you just need it to build success, and you don't care it build in the original folder(qxmpp instead of qxmpp/build)

try
```
git clone https://github.com/qxmpp-project/qxmpp.git
cd qxmpp
cmake
make
```

olesalscheider:
Please try if the first commit in https://github.com/qxmpp-project/qxmpp/pull/145 fixes your problem.

jlaine:
I am unable to reproduce this issue, could you try with the latest code and let us know if the error still occurs?

vasilyrud:
First commit in https://github.com/qxmpp-project/qxmpp/pull/145 fixes the compilation problem. (commit 98753bbaed56f5ad64e725c1164b6fe109447944)

[...]

PR #145, opened 2018-09-06
CMake: Add binary paths for tests to the include paths

olesalscheider:
This PR contains three small fixes for the CMake build system.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -12,6 +12,7 @@ include_directories(${PROJECT_SOURCE_DIR}/src/base)
 include_directories(${PROJECT_SOURCE_DIR}/src/client)
 include_directories(${PROJECT_SOURCE_DIR}/src/server)
 include_directories(${PROJECT_BINARY_DIR}/src/base)
+include_directories(${CMAKE_CURRENT_BINARY_DIR})
 
 add_simple_test(qxmpparchiveiq)
 add_simple_test(qxmppbindiq)
--- a/tests/qxmpptransfermanager/CMakeLists.txt
+++ b/tests/qxmpptransfermanager/CMakeLists.txt
@@ -1,3 +1,4 @@
+include_directories(${CMAKE_CURRENT_BINARY_DIR})
 add_executable(tst_qxmpptransfermanager tst_qxmpptransfermanager.cpp tst_qxmpptransfermanager.qrc)
 add_test(tst_qxmpptransfermanager tst_qxmpptransfermanager)
 target_link_libraries(tst_qxmpptransfermanager Qt5::Test qxmpp)
--- a/tests/qxmpputils/CMakeLists.txt
+++ b/tests/qxmpputils/CMakeLists.txt
@@ -1,3 +1,4 @@
+include_directories(${CMAKE_CURRENT_BINARY_DIR})
 add_executable(tst_qxmpputils tst_qxmpputils.cpp tst_qxmpputils.qrc)
 add_test(tst_qxmpputils tst_qxmpputils)
 target_link_libraries(tst_qxmpputils Qt5::Test qxmpp)


jlaine:
I don't understand all of the changes, but if we could confirm it fixes #141 I'm OK with you merging it.

It feels as though we need to set up appveyor to run automatic win32 builds for us.

jlaine:
Actually wait I'm not able to reproduce the issue with building out of source directory. I'll cherry-pick c++-11 change but I don't understand the rest of the changes.

jlaine:
> Why do you need this change at all?
> GCC use -std=gnu++14 by default since version 6.x.
> See: https://gcc.gnu.org/gcc-6/changes.html

Hm that explains why I'm not getting a warning locally, only on travis (Ubuntu/xenial). However gcc 4.8 is still a supported platform by Qt:

http://doc.qt.io/qt-5/supported-platforms.html

olesalscheider:
Yes, the default standard is different for different versions of gcc. Also I'm not sure if CMake might set a default because I also get this warning with GCC 8.2.
So it is definitely better to be explicit about the standard that we want to use and support.

olesalscheider:
> Actually wait I'm not able to reproduce the issue with building out of source directory. I'll cherry-pick c++-11 change but I don't understand the rest of the changes.

I don't have the issue on my computer but I could reproduce it on an Ubuntu 16.04 system. There it is fixed by the first commit in this PR.

The last commit is important for projects using qxmpp: Without it, you can only try to find qxmpp (`find_package(qxmpp)`) but you cannot specify a minimum version (e.g. `find_package(qxmpp 0.9.4)`).

jlaine:
For the build failure : instead of blindly hoping it fixes something I'd rather we reproduce the problem with our travis setup (which uses Ubuntu/xenial!).

The second change seems like it's unrelated, split it out to its own PR, let CI run an we can merge it.

olesalscheider:
Our current travis setup does not build out-of-source. I guess that's why the problem does not occur there. Maybe we should change that.

jlaine:
I've added the c++-11 change to master.

I've also submitted PR #151 to see whether Travis reproduces the build error.

CMakeLists.txt
```diff
@@ -11,6 +11,8 @@ mark_as_advanced(VERSION_MAJOR VERSION_MINOR VERSION_PATCH VERSION_STRING)
 
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${PROJECT_SOURCE_DIR}/cmake/modules")
 
+set(CMAKE_CXX_STANDARD 11)
```
0xd34df00d: You also probably want `CMAKE_CXX_STANDARD_REQUIRED` set to true (cf. https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD_REQUIRED.html#prop_tgt:CXX_STANDARD_REQUIRED).

olesalscheider:
I have split the PR and submitted new PRs for the other changes.
This PR now only contains the fix for older versions of CMake.

jlaine:
As this is specifically to support older cmake versions, is there a way we can add such old cmake versions to our CI build matrix? Otherwise it's bound to break again somewhere down the line.

olesalscheider:
I'm not sure. We would have to use the native CMake version inside the CI instead of the one provided by travis.

It is easy to reproduce the issue with Gitlab CI because it allows to run the CI in an unmodified xenial docker container: https://gitlab.com/olesalscheider/qxmpp/-/jobs/102949969

But I guess we cannot select the docker container with travis. The only other thing I can think of is that there might be multiple versions of CMake in the travis container: The native from xenial and a newer one from travis. In this case we could maybe call the original version...

olesalscheider:
I have fixed the issue in https://github.com/qxmpp-project/qxmpp/pull/155

PR #147, opened 2018-09-07
CMake: Build examples and documentation by default

olesalscheider:


CMakeLists.txt
```diff
@@ -18,8 +18,8 @@ set(CMAKE_AUTORCC ON)
 include(GNUInstallDirs)
 
 option(BUILD_TESTS "Build tests." ON)
-option(BUILD_DOCUMENTATION "Build API documentation." OFF)
-option(BUILD_EXAMPLES "Build API documentation." OFF)
+option(BUILD_DOCUMENTATION "Build API documentation." ON)
+option(BUILD_EXAMPLES "Build API documentation." ON)
```
jlaine: I was actually OK with not building the documentation by default, as this drags an extra dependency on Doxygen, which may not be trivial to satisfy on non-unix platforms (especially getting in on the PATH..)

olesalscheider:
Ok, yes that makes sense. I have updated the patch.

PR #151, opened 2018-09-18
[travis] build qxmpp out-of-source

jlaine:


jlaine:
Let's see if this reproduces #141

jlaine:
Well, "unfortunately" the build passes..

tehnick:
@jlaine Sorry, I accidentally have pushed my changes into your qxmpp fork instead of mine.

tehnick:
@olesalscheider As we see build in Ubuntu Xenial environment on Travis CI is successfull too.

tehnick:
Just a side note: the cross-compilation of qxmpp with default configuration options using mingw is currently broken. My CI system checks master branch of qxmpp project. You may look at last build logs here: https://tehnick.net/sibuserv/projects/qxmpp/v0.9.3-49-g09560c9

olesalscheider:
That's interesting. In the CI, the CMake version is 3.9.2. On the system I tested the CMake version is 3.5.1 which is also the default for xenial: https://packages.ubuntu.com/xenial/cmake

Do you have an idea why travis has the newer version? But this might explain the different behaviour.

tehnick:
https://docs.travis-ci.com/user/languages/cpp/#cmake

tehnick:
https://docs.travis-ci.com/user/build-environment-updates/2017-11-08/

jlaine:
Hm. If the problem is just with old cmakes I'm tempted to document it and just move on..

olesalscheider:
Well, it's a problem with the CMake version shipped with xenial (and probably other distributions). If you do not want to support that version users of xenial have to update it manually. The documentation of CMake changed and the old version contains a hint that you should add the build directory to the include path. Newer versions do this automatically.

So this is the right fix if you want to support the older versions...

tehnick:
> https://github.com/qxmpp-project/qxmpp/pull/151#issuecomment-422385211

That info is outdated. With current master all builds and static code analyzer checks are passed fine: https://tehnick.net/sibuserv/projects/qxmpp/v0.9.3-56-g34b6a75

PR #155, opened 2018-09-30
travis: Only use system binaries

olesalscheider:
With this change we will use the CMake version provided by Ubuntu Xenial and not the updated one from Travis.


jlaine:
Ah very nice, now the build fails :)

PR #159, opened 2018-10-14
Implement XEP-0352: Client State Indication

LNJ2:
Due to inactivity of the author in PR #87, I opened this pull request with the rebased changes and some modifications.


src/client/QXmppClient.cpp
```diff
@@ -320,6 +324,28 @@ bool QXmppClient::isConnected() const
...
+    return d->isActive;
+}
+
+/// Sets the client state as described in to XEP-0352
+
+void QXmppClient::setActive(bool active)
+{
+    if (active != d->isActive && d->stream->isClientStateIndicationEnabled()) {
+        d->isActive = active;
+        QString packet("<%1 xmlns='%2'/>");
```
0xd34df00d: This is probably a microoptimisation given that network is involved, but `ns_csi` is statically known, so maybe it's worth pasting that into `packet` straight away?

0xd34df00d: This is probably a microoptimisation given that network is involved, but `ns_csi` is statically known, so maybe it's worth pasting that into packet straight away?

src/client/QXmppClient.cpp
```diff
@@ -320,6 +324,28 @@ bool QXmppClient::isConnected() const
...
+        if (active)
+            d->stream->sendData(packet.arg("active", ns_csi).toUtf8());
```
0xd34df00d: I'd just
```c++
d->stream->sendData(packet.arg(active ? "active" : "inactive", ns_csi).toUtf8());
```
but that might hinder readability a bit, so that's up to you.

0xd34df00d:
But let's hear from @jlaine too.

src/client/QXmppClient.cpp
LNJ2: I like your suggestion, I'll use it. :smiley:

jlaine:
On the whole this looks very good, but could we have some tests which exercise the new QXmppClient methods? Once that's in, this PR can be merged

LNJ2:
If yes, then I don't really know how to test the new methods (isActive/setActive) since they require an established XMPP connection. I could create a local QXmppServer and use that for it.

jlaine:
Hm, you are right there is no explicit test for a basic QXmppClient per se. The test for QXmppTransferManager does however connect two clients to a local server. I think I'll merge this PR do you mind following up with another PR to add a test?

LNJ2:
I just squashed/rebased the commits.

@jlaine You mean adding tests for the QXmppClient, right? I can do that. :)

LNJ2:
I started writing a test for the QXmppClient/CSI, but I can't enable CSI support in the QXmppServer, so I can't test the behaviour with enabled CSI. I could add a method to the QXmppServer to enable the CSI feature (so the QXmppIncomingClient can check whether to include the stream feature), but that is probably not the ideal way. I also don't think adding a QXmppServerExtension (and `QXmppServerExtension::streamFeatures()`) were enough, because CSI should filter and cache the packets sent to the client (which probably needs changes in the QXmppIncomingClient).

jlaine:
A server extension sounds like the right way to go at least you can check that feature detection works. We can refine tests in another PR

src/server/QXmppIncomingClient.h
```diff
@@ -41,7 +41,8 @@ class QXMPP_EXPORT QXmppIncomingClient : public QXmppStream
     Q_OBJECT
 
 public:
-    QXmppIncomingClient(QSslSocket *socket, const QString &domain, QObject *parent = 0);
+    QXmppIncomingClient(QSslSocket *socket, const QString &domain, bool csiEnabled = false,
+                        QObject *parent = 0);
```
LNJ2: Warning: This change breaks compatibility. Alternative were another constructor and defaulting `csiEnabled` to false with the current one.

LNJ2:
Currently the problem in the test is, that the `<inactive/>` tag is sent by the client but not received by the QXmppServer (other stanzas are received).

LNJ2:
I removed state checks from server side for now. At least there is a basic test that will check the client's behaviour, now. By adding this test I actually also found a bug (QXmppStreamFeatures had enabled CSI with default constructor).

LNJ2:
@jlaine Is the test I added enough and are the other changes ok? Also, codecov comment seems to be buggy: it displays that I was removing tests for header files. (it's ok on their website: https://codecov.io/gh/qxmpp-project/qxmpp/pull/159)

jlaine:
OK, the test is well and truly broken on Windows. I'd like to just merge the code changes and rework the test but not sure how to proceed.

LNJ2:
> Edit 2: OK I see the problem, the server extension mechanism has no provision for influencing stream features. Would it make sense to add this?

If this was a question to me: I think it would. There are other XEPs which require to include additional stream features and it would make the core more extensible (could avoid forks in some cases).

LNJ2:
Just FYI it's actually working:
https://git.kaidan.im/kaidan/kaidan/commit/cff4338a479474599f49eea8d52904e163eef876

PR #168, opened 2018-12-06
Use QSysInfo for client OS in VersionManager

lnjX:
The QXmppVersionManager will use QSysInfo in Qt 5.4 or later to
determine the client's OS, so it will also contain the OS version or
some codenames.

### DIFF
--- a/src/client/QXmppDiscoveryManager.cpp
+++ b/src/client/QXmppDiscoveryManager.cpp
@@ -48,7 +48,11 @@ QXmppDiscoveryManager::QXmppDiscoveryManager()
 {
     d->clientCapabilitiesNode = "https://github.com/qxmpp-project/qxmpp";
     d->clientCategory = "client";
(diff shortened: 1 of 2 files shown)


olesalscheider:
This looks good to me. Maybe it would also make sense to set the client type (https://github.com/qxmpp-project/qxmpp/blob/83d311bf66dad95ea4e727ceffc89771d630b123/src/client/QXmppDiscoveryManager.cpp#L51) based on the operating system (QSysInfo::productType)?

jlaine:
@olesalscheider I'm not convinced QSysInfo::productType makes things easier

olesalscheider:
You mean instead of the #ifdefs that @lnjX pushed yesterday? I think they work as well and there is no reason to prefer QSysInfo::productType.
But in general it is a good idea to default to the "phone" type on a mobile OS. Don't you think so?

PR #170, opened 2018-12-29
Implement XEP-0308: Last Message Correction

lnjX:


jlaine:
As usual, good stuff with docstrings and a test!

If you could just add a test that exercises the setter for "replaceId" I'd be happy to merge this.

jlaine:
Thanks!

PR #172, opened 2019-01-02
README: Link to supported XEPs in wiki

lnjX:


jlaine:
While I'm not absolutely opposed to this, this seem to duplicate the information from:

https://github.com/qxmpp-project/qxmpp/blob/master/doc/xep.doc

Why do we prefer the wiki?

lnjX:
Oh, I actually wasn't aware of that list, because there's no link to. Then I'd prefer the page in the documentation. Only problems I see: afaik we can't directly link to the html version of the page from the README, because the QXmpp version number is in the [URL](http://doc.qxmpp.org/qxmpp-0.9.1/xep.html), currently the page is outdated and it's always from the latest stable release (we could solve that by doing frequent releases :D).

jlaine:
OK, could you replace this PR by a PR which updates doc/xep.doc for a start? In the future we need to make sure PRs which add support for additional XEPs include the relevant change to doc/xep.doc

PR #174, opened 2019-01-03
Implement XEP-0369/XEP-0405 (MIX): IQ queries

lnjX:
This implements all used IQ queries of XEP-0369: Mediated Information
eXchange (MIX) (v0.14.0) and XEP-0405: Mediated Information eXchange (MIX):
Participant Server Requirements (v0.3.0), including unit tests.

Happy reviewing! :D


lnjX:
Do you have any idea why some of the travis builds fail (I'm not having that error locally with gcc 8.2)? Didn't I do exactly the same as in the other tests?

lnjX:
... now it's really fixed (there was a `(` too much)

lnjX:
Now also added a test for `::isMixIq()`, should be ready now.

lnjX:
ping! :)
It were nice if you could review & merge the three MIX pull requests soon: It were also easier for me to work on the MixManager then.

src/base/QXmppMixIq.cpp
```diff
@@ -0,0 +1,181 @@
...
+    m_actionType = (QXmppMixIq::Type) MIX_ACTION_TYPES.indexOf(child.tagName());
+
+    if (child.namespaceURI() == ns_mix_pam) {
+        if (child.hasAttribute("channel"))
+            m_jid = child.attribute("channel");
+
+        child = child.firstChildElement();
+    }
+
+    if (child.isNull() || child.namespaceURI() != ns_mix)
```
jlaine: I'm finding this condition a bit confusing, maybe turn it into a "positive" condition?

if (!child.isNull() && child.namespaceURI() == ns_mix) {
  // do something with the child
}

src/base/QXmppMixIq.cpp
```diff
@@ -0,0 +1,181 @@
...
+    const QDomElement& child = element.firstChildElement();
+    return !child.isNull() && (child.namespaceURI() == ns_mix
+           || child.namespaceURI() == ns_mix_pam);
+}
+
+void QXmppMixIq::parseElementFromChild(const QDomElement& element)
+{
+    QDomElement child = element.firstChildElement();
+    // determine action type
+    m_actionType = (QXmppMixIq::Type) MIX_ACTION_TYPES.indexOf(child.tagName());
```
jlaine: Will this blow up if child is null?

src/base/QXmppMixIq.cpp
lnjX: Done :)

src/base/QXmppMixIq.cpp
lnjX: No, it won't. I also added a test for the empty case.

PR #175, opened 2019-01-05
Implement MIX message, presence and roster extensions

lnjX:


src/base/QXmppRosterIq.cpp
```diff
@@ -253,6 +280,34 @@ void QXmppRosterIq::Item::setSubscriptionTypeFromStr(const QString& type)
...
+/// Sets whether this is a MIX channel.
+
+void QXmppRosterIq::Item::setIsMixChannel(bool isMixChannel)
+{
+    m_isMixChannel = isMixChannel;
+}
+
+/// Returns the participant id for this MIX channel.
+
+QString QXmppRosterIq::Item::participantId() const
```
jlaine: Should this be `mixParticipantId` to keep everything MIX-related with a consistent prefix?

lnjX: done

src/base/QXmppRosterIq.h
```diff
@@ -86,6 +93,9 @@ class QXMPP_EXPORT QXmppRosterIq : public QXmppIq
         // can be subscribe/unsubscribe (attribute "ask")
         QString m_subscriptionStatus;
         QSet<QString> m_groups;
+        // XEP-0405: Mediated Information eXchange (MIX): Participant Server Requirements
+        bool m_isMixChannel = false;
```
jlaine: This breaks ABI compatibility. If we're going to do this let's at least future-proof it by moving everything to a d-pointer.

src/base/QXmppRosterIq.h
```diff
@@ -108,6 +122,8 @@ class QXMPP_EXPORT QXmppRosterIq : public QXmppIq
     QList<Item> m_items;
     // XEP-0237 Roster Versioning
     QString m_version;
+    // XEP-0405: Mediated Information eXchange (MIX): Participant Server Requirements
```
jlaine: Same as above: let's move everything behind a d-pointer.

src/base/QXmppRosterIq.h
tehnick: Useful link about d-pointer and about ABI compatibility in common: 
https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B

And I would prefer the separate commits for this task. For not mixing this with changes related to providing of new features in this PR...

src/base/QXmppRosterIq.h
lnjX: I can do that after https://github.com/qxmpp-project/qxmpp/pull/189 has been merged.

lnjX: done

src/base/QXmppRosterIq.h
lnjX: done

lnjX:
If the ABI break is a problem for you, you could maybe merge everything into a new develop branch or something like this.

tehnick:
I do not think that ABI breaks may be a problem for now. Because we only at the beginning of a way of introducing the practice of maintaining of stable ABI in qxmpp.

lnjX:
Is there anything because this pull request can't be merged?

lnjX:
@jlaine Please come back from your reviewing holidays :)

lnjX:
Updated implementation to latest XEP version.

CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ cmake_minimum_required(VERSION 3.2)
 set(VERSION_MAJOR 1)
 set(VERSION_MINOR 0)
 set(VERSION_PATCH 1)
-set(SO_VERSION 1)
+set(SO_VERSION 2)
```
jlaine: This looks unrelated to this PR

jlaine:
> @jlaine Please come back from your reviewing holidays :)

I wish someone else would take part in the review process.. It's a bit daunting when there are 20 PRs to review!

CMakeLists.txt
lnjX: Yes, that's the problem.

lnjX:
@jlaine Thanks for merging! :)

> I wish someone else would take part in the review process.. It's a bit daunting when there are 20 PRs to review!

I'd like to do so, but I also can't review my own pull requests.

PR #179, opened 2019-01-05
Implement XEP-0369 (MIX): Info/Participant node items

lnjX:
This implements the pubsub items for the MIX participants and info node.

https://xmpp.org/extensions/xep-0369.html#participants-node
https://xmpp.org/extensions/xep-0369.html#info-node

I'm just not entirely sure whether the current pubsub item principle is such a good idea. But it currently "works".


src/base/QXmppMixItem.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+    QXmppDataForm::Field contactsField;
+    contactsField.setKey("Contact");
+    contactsField.setValue(m_contactJids);
+    contactsField.setType(QXmppDataForm::Field::JidMultiField);
+    fields << contactsField;
+
+    form.setFields(fields);
+
+    // FIXME: this is too complicated; maybe don't use QXmppElement in QXmppPubSubItem?
```
lnjX: Another solution would be to use QXmppElement for everything; parsing and serialization. But this and using a proper JID class would probably both be tasks for "QXmpp 2.0", although it were theoretically possible to keep compatibility by creating all methods with both parameter types.

jlaine: I think I'd rather see QXmppElement die that expanding its usage

jlaine:
I really have no knowledge of MIX.. All I can say is this PR's code style is consistent with QXmpp's API, it does a great job at raising the code coverage.

There are zero documentation strings though, could we improve that?

lnjX:
> There are zero documentation strings though, could we improve that?

Done :)

jlaine:
It seems `isMixChannelInfo` is lacking a test. I can't remember how hard it is to set up such a test?

lnjX:
> It seems isMixChannelInfo is lacking a test. I can't remember how hard it is to set up such a test?

Good that you said this, `isMixParticipantItem()` was still missing (and `isMixChannelInfo()` was outdated :man_shrugging: ).

lnjX:
Should be ready now! :)

jlaine:
Where is this going to be used? AFAICT these "items" are not IQs in themselves, so what elements carry them?

lnjX:
These items are pubsub items and they are used when the channel description is changed/published (pubsub) and also when a user joins a channel (via. IQ) a pubsub notification is sent to all subscribers of the participants pubsub node. The pubsub notifications are message extensions.

PR #183, opened 2019-01-16
[sasl] add support for SCRAM-SHA-1 and SCRAM-SHA-256

jlaine:
I have been able to test SCRAM-SHA-1 against the jabber.at server, does anyone know one which supports SCRAM-SHA-256?


jlaine:
I have been able to test SCRAM-SHA-1 against the jabber.at server, does anyone know one which supports SCRAM-SHA-256?

PR #187, opened 2019-01-17
[sasl] order mechanisms to prefer the most secure

jlaine:
The previous logic was:

- use the preferred SASL mechanism if available
- otherwise use the first supported mechanism offered by the server

However RFC 6120, section 6.3.3 states:

"The initiating entity MUST maintain its own preference order independent
of the preference order of the receiving entity."

The new logic is:

- order our supported mechanisms from most secure to least secure
- if the user sets QXmppConfiguration::saslMechanism, put it first
- use the best mechanism supported by the server

### DIFF
--- a/src/base/QXmppSasl.cpp
+++ b/src/base/QXmppSasl.cpp
@@ -272,8 +272,8 @@ QXmppSaslClient::~QXmppSaslClient()
 
 QStringList QXmppSaslClient::availableMechanisms()
 {
-    return QStringList() << "PLAIN" << "DIGEST-MD5" << "ANONYMOUS"
-                         << "SCRAM-SHA-1" << "SCRAM-SHA-256"
+    return QStringList() << "SCRAM-SHA-256" << "SCRAM-SHA-1" << "DIGEST-MD5"
+                         << "PLAIN" << "ANONYMOUS"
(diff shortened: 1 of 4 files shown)


jlaine:
@lnjX @olesalscheider @0xd34df00d @tehnick I would appreciate your feedback on this change, as it changes the behaviour of QXmpp. I believe it's better than before, but still it's worth discussing.

I wasn't too sure where to put the proprietary mechanisms as I don't know how to assess their relative strength.

olesalscheider:
This looks good to me. I'm also not sure about the proprietary mechanisms. But after a quick search it seems that oauth transmits the plain credentials over https to the authorization server. That's probably as secure as PLAIN if the xmpp connection uses TLS.

But I guess in practise it does not matter that much: Most servers won't offer the proprietary mechanisms. And the ones that do won't accept anything else, I guess?

jlaine:
OK, I'll merge as-is then and we can always iterate if we find a better logic. I have been trying the new SCRAM-SHA-x support against some public servers, and we really do need this new logic, otherwise we end up selecting PLAIN since our previous hard-coded DIGEST-MD5 preference is usually not offered.

lnjX:
I also think this can only improve things. :)

PR #188, opened 2019-01-19
Implement XEP-0363: HTTP File Upload: Request IQs

lnjX:
This implements the IQs for requesting and receiving upload slots as
defined by XEP-0363: HTTP File Upload [v0.9.0].


doc/xep.doc
```diff
@@ -46,5 +46,6 @@ Ongoing:
 - XEP-0009: Jabber-RPC (API is not finalized yet)
 - XEP-0060: Publish-Subscribe (Only basic IQ implemented)
 - XEP-0077: In-Band Registration (Only basic IQ implemented)
+- XEP-0363: HTTP File Upload [v0.9.0]
```
jlaine: What does the version refer to?

src/base/QXmppHttpUploadIq.cpp
```diff
@@ -0,0 +1,195 @@
...
+ *
+ */
+
+#include <QDomElement>
+#include <QMimeDatabase>
+
+#include "QXmppHttpUploadIq.h"
+#include "QXmppConstants_p.h"
+
+QString QXmppHttpUploadRequestIq::fileName() const
```
jlaine: Could we have some doc strings on the members please?

src/base/QXmppHttpUploadIq.h
```diff
@@ -0,0 +1,97 @@
...
+    static bool isHttpUploadRequestIq(const QDomElement &element);
+
+protected:
+    /// \cond
+    void parseElementFromChild(const QDomElement &element);
+    void toXmlElementFromChild(QXmlStreamWriter *writer) const;
+    /// \endcond
+
+private:
+    QString m_fileName;
```
jlaine: I tried to put members into an opaque structure wherever possible to preserve ABI in case of future changes, does this make sense here too?

src/base/QXmppHttpUploadIq.h
```diff
@@ -0,0 +1,97 @@
...
+class QXMPP_EXPORT QXmppHttpUploadRequestIq : public QXmppIq
+{
+public:
+    QString fileName() const;
+    void setFileName(const QString &filename);
+
+    qint64 size() const;
+    void setSize(const qint64 &size);
+
+    QMimeType contentType() const;
```
jlaine: This is interesting, I had forgotten we gained a QMimeType class in Qt 5!

doc/xep.doc
lnjX: It's the XEP's version, so we can easily check if an implementation is outdated.

lnjX: I also added the version number in the header files, maybe that's a bit too much?

src/base/QXmppHttpUploadIq.h
jlaine: I think we should have a private class, yes. I didn't expect roster IQs to change either, but now it turns out your other PR is going to break ABI..

lnjX: Then we should also use a d-pointer for `QXmppMixIq` (I could imagine changes there as the protocol isn't finished yet and hasn't been practically used).

lnjX:
Now with a d-pointer.

src/base/QXmppHttpUploadIq.cpp
```diff
@@ -0,0 +1,262 @@
...
+/// Returns the file's size in bytes.
+
+qint64 QXmppHttpUploadRequestIq::size() const
+{
+    return d->size;
+}
+
+/// Sets the file's size in bytes.
+
+void QXmppHttpUploadRequestIq::setSize(const qint64 &size)
```
jlaine: Passing integers by reference is .. unusual? :) I think a plain setSize(qint64 size) would be fine

src/base/QXmppHttpUploadIq.cpp
```diff
@@ -0,0 +1,262 @@
...
+    }
+}
+
+/// Returns whether the URLs (put and get) are both HTTPS URLs. It will return
+/// false, if the URLs are non-HTTPS URLs or not set. HTTP URLs are not
+/// XEP-compliant, but are used in some cases.
+
+bool QXmppHttpUploadSlotIq::hasHttpsUrls() const
+{
+    return d->getUrl.toString().startsWith("https://") &&
```
jlaine: Do we need this method? If we absolutely do, why not use url.scheme() ?

src/base/QXmppHttpUploadIq.cpp
```diff
@@ -0,0 +1,262 @@
...
+    }
+    return false;
+}
+
+/// \cond
+void QXmppHttpUploadSlotIq::parseElementFromChild(const QDomElement &element)
+{
+    QDomElement slot = element.firstChildElement("slot");
+    QDomElement put = slot.firstChildElement("put");
+    d->getUrl = QUrl::fromEncoded(slot.firstChildElement("get").attribute("url").toUtf8());
```
jlaine: This is slightly confusing, why are we not handling get / put in a symetric fashion? Also is either of these URLs optional?

src/base/QXmppHttpUploadIq.cpp
```diff
@@ -0,0 +1,262 @@
...
+void QXmppHttpUploadSlotIq::setGetUrl(const QUrl &getUrl)
+{
+    d->getUrl = getUrl;
+}
+
+/// Returns a map of header fields (header name -> content) that need to be
+/// included in the PUT (upload) request. This won't contain any other fields
+/// than: "Authorization", "Cookie" or "Expires".
+
+QMap<QString, QString> QXmppHttpUploadSlotIq::headerFields() const
```
jlaine: Would putHeaderFields clarify that these headers only apply to PUT?

src/base/QXmppHttpUploadIq.cpp
lnjX: Oh that was an accident. :joy:

src/base/QXmppHttpUploadIq.cpp
lnjX: A valid slot needs both URLs to be set. But you're right, the behaviour/https-checking can be done in the manager (and yes `.scheme()` were better of course :smile:).

src/base/QXmppHttpUploadIq.cpp
lnjX: Let's call it just `putHeaders` then.

jlaine: Even better

src/base/QXmppHttpUploadIq.cpp
```diff
@@ -0,0 +1,253 @@
...
+/// request. All fields other than "Authorization", "Cookie" or "Expires" will
+/// be ignored.
+
+void QXmppHttpUploadSlotIq::setPutHeaders(const QMap<QString, QString> &putHeaders)
+{
+    d->putHeaders.clear();
+    for (QString &name : putHeaders.keys()) {
+        if (name == "Authorization" || name == "Cookie" || name == "Expires") {
+            // newlines must be removed
+            QString value = putHeaders[name];
```
jlaine: I'm not convinced "removing" newlines is the best option. Also I expect \r isn't accepted either?

Wouldn't failing loudly be more appropriate?

src/base/QXmppHttpUploadIq.cpp
```diff
@@ -0,0 +1,253 @@
...
+/// server's job).
+
+void QXmppHttpUploadRequestIq::setFileName(const QString &fileName)
+{
+    d->fileName = fileName;
+}
+
+/// Returns the file's size in bytes.
+
+qint64 QXmppHttpUploadRequestIq::size() const
```
jlaine: I'm sorry for pointing out these things one by one.. but why fileName() vs size() ? I would expect either name() and size() or fileName() and fileSize().

QXmppTransferFileInfo use name() and size()

lnjX: Yeah right, I took the XML attr. names.

lnjX: Is that ok or should I better use consistent names?

src/base/QXmppHttpUploadIq.cpp
lnjX: I'm not sure, but maybe Qt is already doing this anyway, so we might don't need to remove newlines on our own.

jlaine:
Could you rebase this please? Also, will there be a manager to handle these IQs?

lnjX:
Done.

Yes there will be a manager or even two: One for requesting/receiving the slots from the server and one for actually uploading a file from disk.

tehnick:
Thank you!

> I just didn't want to add too much to one pull request, so reviewing is at least a bit easier.

And this really helps.

PR #189, opened 2019-01-22
roster iq: Make use of d-pointer

lnjX:


lnjX:
Is this ok?

lnjX:
Moved into #175.

PR #191, opened 2019-01-29
Add not authorized condition for stream error

henry61024:
Support `not-authorized` condition for `stream:error`.

This allow us to distinguish `stream:error` conditions more specifically.

Please refer to https://xmpp.org/rfcs/rfc6120.html#streams-error-conditions-not-authorized.

### DIFF
--- a/src/client/QXmppOutgoingClient.cpp
+++ b/src/client/QXmppOutgoingClient.cpp
@@ -563,7 +563,7 @@ void QXmppOutgoingClient::handleStanza(const QDomElement &nodeRecv)
 
         if (!nodeRecv.firstChildElement("conflict").isNull())
             d->xmppStreamError = QXmppStanza::Error::Conflict;
-        else if (!nodeRecv.firstChildElement("not-authorized").isNull() || !nodeRecv.firstChildElement("bad-auth").isNull())
+        else if (!nodeRecv.firstChildElement("not-authorized").isNull())
             d->xmppStreamError = QXmppStanza::Error::NotAuthorized;
         else
             d->xmppStreamError = QXmppStanza::Error::UndefinedCondition;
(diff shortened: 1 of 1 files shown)


jlaine:
Could you add some context about how this condition is encountered? Usually the "not-authorized" comes as part of the SASL exchange.

Also: do we really need the "bad-auth" case you added? It was added in the SASL handling as a workaround for faulty servers, I'd like to know for sure we **need** it at the stream level before adding it too.

henry61024:
Because we want to mange our user accounts with features such as remove/lockout users.

When a user has been remove/lockout, xmpp server([Openfire](https://www.igniterealtime.org/projects/openfire/)) will send a `not-authorized` package in order to kickoff the unauthorized user. So our clients also need to support the `not-authorized` package.

About the `bad-auth`: I'm just referring to the error condition in

 https://github.com/qxmpp-project/qxmpp/blob/ba93161573503c666df1e6ba141e157ecdc6fd27/src/client/QXmppOutgoingClient.cpp#L614

lnjX:
Don't we need to add all the other error cases, too?

src/client/QXmppOutgoingClient.cpp
```diff
@@ -563,6 +563,8 @@ void QXmppOutgoingClient::handleStanza(const QDomElement &nodeRecv)
 
         if (!nodeRecv.firstChildElement("conflict").isNull())
             d->xmppStreamError = QXmppStanza::Error::Conflict;
+        else if (!nodeRecv.firstChildElement("not-authorized").isNull() || !nodeRecv.firstChildElement("bad-auth").isNull())
```
lnjX: The 'bad-auth' thing was a workaround for (old) prosody servers. Could you remove that. I think we can merge it then.

henry61024: Done. Fixed in b2bcfd9

PR #192, opened 2019-02-01
MessageReceiptManager: Don't accept receipts from other resources

lnjX:
There were some problems with buggy clients leading to the problem that some
messages were already marked as received, even though just another
resource of the used account had received the message.

Here is an example:
```
[outgoing]:
<message id="tH9OkRw"
         to="42@example.com"
         from="lnj@kaidan.im/kaidan.PR29"
         type="chat">
    <body>test</body>
    <n1:request xmlns:n1="urn:xmpp:receipts"/>
</message>

[incoming]:
<message to="lnj@kaidan.im/kaidan.PR29"
         from="lnj@kaidan.im/dino.dc02d539"
         id="410b33c3-1cd3-433e-8699-74a7583c2560">
    <n1:received xmlns="urn:xmpp:receipts" id="tH9OkRw"/>
</message>
```
Here the other client "dino.dc02d539" sent a `<received/>` tag back, although
the author had the same bare JID. To avoid that we need to
ignore incoming received messages from our bare JID.


jlaine:
Could we have a unit test capturing this please?

lnjX:
Done

jlaine:
Beautiful thank you!

PR #194, opened 2019-03-09
Implement XEP-0392: Consistent Color Generation

lnjX:
This adds methods to generate a color for a JID (or nickname or whatever).
It also supports the color corrections for color vision deficiencies.
Only background color correction is not covered by this, but may be
added later.

It uses the hsluv-c library to convert the hue/saturation/lightness
values to RGB colors. hsluv-c is licensed under MIT and is included now.

### DIFF
--- a/doc/xep.doc
+++ b/doc/xep.doc
@@ -41,6 +41,7 @@ Complete:
 - XEP-0313: Message Archive Management
 - XEP-0319: Last User Interaction in Presence
 - XEP-0352: Client State Indication
+- XEP-0392: Consistent Color Generation (v0.6.0)
 
 Ongoing:
 - XEP-0009: Jabber-RPC (API is not finalized yet)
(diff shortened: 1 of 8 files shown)


0xd34df00d:
What do you think about using `QColor` to initialize it with the HSL components and then using `toRgb()`? This way the dependency on `hsluv-c` could be avoided (and, personally, as a Gentoo qxmpp maintainer, I would rather avoid bundling libraries).

lnjX:
QColor is from QtGui which would add a new dependency, that's why I used hsluv-c. I personally think it's ok to bundle hsluv-c, because it's very small.

lnjX:
Also, I can only find HSL support in QColor, which is afaik different than HSLuv.

lnjX:
I can't seem to find the error in the AppVeyor builds, could someone tell me what's the problem there? :)

tehnick:
> I personally think it's ok to bundle hsluv-c, because it's very small.

Even if including of this library into source tree will be accepted by @jlaine I would like to see it in `3rdparty` or `thirdparty` subdirectory.

@jlaine What do you think?

PR #195, opened 2019-04-27
Implement XEP-0382: Spoiler messages

lnjX:
This adds parsing and serialization of spoilers in the QXmppMessage class.


jlaine:
This needs a rebase now that MIX changes have landed

lnjX:
done

PR #199, opened 2019-04-28
Implement XEP-0380: Explicit Message Encryption

lnjX:
This adds parsing and serialization for XEP-0380: Explicit Message
Encryption in version 0.2.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ QXmpp 1.0.1 (UNRELEASED)
 New features:
  - Add support for SCRAM-SHA-1 and SCRAM-SHA-256 (#183, @jlaine)
  - Add XEP-0367: Message Attaching (v0.3.0) (#196, @lnjX)
+ - Add XEP-0380: Explicit Message Encryption (v0.3.0) (#199, @lnjX)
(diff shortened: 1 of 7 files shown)


jlaine:
I'm not too sure about the multiple ways of setting the encryption method. Does it even make sense to support "custom" encryption methods?

lnjX:
> I'm not too sure about the multiple ways of setting the encryption method. Does it even make sense to support "custom" encryption methods?

The XEP says we need to support custom methods and I think this is useful, so if there is a new encryption method (or you do your own), you don't need to wait that it's added in QXmpp, QXmpp has a new release and that new release is distributed everywhere, so you can finally use it in your project.

> Also, will we be supporting the actual encryption process?

Yes, we (Kaidan developers) are going to support OMEMO and we will also add everything for that in QXmpp. We won't support Legacy OpenPGP and OTR, though (not sure about OX).

> The trend of the recent PRs seems to be to add plenty of parsing / serialization abilities, but placing the burden of using them on the end user.

This is not what I want and I will add more client managers, for MIX (already 50% finished) and for HTTP File Upload (basically finished). But I don't know what I could do to make e.g. Message Attaching or Spoiler messages easier to use.

PR #200, opened 2019-04-30
Update implementation of XEP-0369 (v0.14.2) and XEP-0405 (v0.4.0)

lnjX:
The changes in the XEP only affect parts we haven't implemented yet, so
updating was rather easy.


src/base/QXmppMixIq.cpp
```diff
@@ -46,10 +46,6 @@ QXmppMixIq::QXmppMixIq()
 {
 }
 
-QXmppMixIq::~QXmppMixIq()
```
jlaine: Does this change ABI?

lnjX: Possibly yes ...
I'll revert this change.

PR #201, opened 2019-04-30
Check TLS-certs against connection host instead of JID domain

lnjX:
Some server setups don't have a certificate for the actual JID domain
(e.g. myserver.org) and only for their local host (e.g. xmpp.myserver.org).


jlaine:
This is a violation of RFC 6120, so no.

PR #204, opened 2019-05-10
Modernize codebase using clang-tidy

JBBgameich:
Using the following checks:
 * modernize-use-nullptr
 * modernize-use-override
 * modernize-use-using
 * modernize-use-bool-literals

This results can be reproduced running the following commands:
```
cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON .
run-clang-tidy-6.0 -header-filter='.*' -checks='-*,modernize-use-nullptr,modernize-use-override,modernize-use-using,modernize-use-bool-literals' -fix
```

[](https://user-images.githubusercontent.com/13609393/57555763-6e469c80-7375-11e9-8d79-5d48d32e7ed0.png)

### DIFF
--- a/examples/example_1_echoClient/example_1_echoClient.h
+++ b/examples/example_1_echoClient/example_1_echoClient.h
@@ -32,8 +32,8 @@ class echoClient : public QXmppClient
     Q_OBJECT
 
 public:
-    echoClient(QObject *parent = 0);
-    ~echoClient();
+    echoClient(QObject *parent = nullptr);
+    ~echoClient() override;
 
 public slots:
     void messageReceived(const QXmppMessage&);
--- a/examples/example_2_rosterHandling/example_2_rosterHandling.h
+++ b/examples/example_2_rosterHandling/example_2_rosterHandling.h
@@ -32,8 +32,8 @@ class xmppClient : public QXmppClient
     Q_OBJECT
 
 public:
-    xmppClient(QObject *parent = 0);
-    ~xmppClient();
+    xmppClient(QObject *parent = nullptr);
+    ~xmppClient() override;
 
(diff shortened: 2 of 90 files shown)


JBBgameich:
The codecov report probably just failed because this touched very old lines of code which are not covered by unit tests.

jlaine:
Is there a way to plug this into a CI check to make sure future additions stay compliant?

tehnick:
> Is there a way to plug this into a CI check to make sure future additions stay compliant?

Ubuntu 16.04 uses clang-tidy-3.8 by default:
https://packages.ubuntu.com/xenial/clang-tidy
Our scripts for Travis CI may be complemented by launch of `run-clang-tidy-8`.

But I strongly do not recommend to use `modernize-use-override` option: it works unstable and often breaks sources (if `run-clang-tidy` is launched with `-fix` option).

JBBgameich:
Does the ci test have to be included in this PR?

lnjX:
I don't think that we really need a CI test for this (would be nice to have of course though).

tehnick:
> Does the ci test have to be included in this PR?

I do not think so. Just a discussion.

tehnick:
Could you revert `modernize-use-override` related changes? Then I'll merge this PR.

lnjX:
> Could you revert modernize-use-override related changes?

Why that? We already broke the binary compatibility before (also after the 1.0).

lnjX:
@tehnick I would like to merge this as it is now. Any particular reason why we should not use override (which generates warnings with current compilers)?

lnjX:
@tehnick I will merge this if you don't have anything to say.

PR #206, opened 2019-05-14
Port away from deprecated qt features

JBBgameich:
This fixes multiple warnings when compiling with Qt 5.13 and will simplify the transition to Qt 6 once it's released.

The next thing to do (in a new PR) is porting away from the deprecated foreach (also known as Q_FOREACH) macro which is unfortunately used 155 times...

### DIFF
--- a/src/client/QXmppOutgoingClient.cpp
+++ b/src/client/QXmppOutgoingClient.cpp
@@ -25,6 +25,7 @@
 #include <QCryptographicHash>
 #include <QNetworkProxy>
 #include <QSslSocket>
+#include <QSslConfiguration>
 #include <QUrl>
 #include <QDnsLookup>
 
@@ -141,8 +142,11 @@ void QXmppOutgoingClientPrivate::connectToHost(const QString &host, quint16 port
     q->info(QString("Connecting to %1:%2").arg(host, QString::number(port)));
 
     // override CA certificates if requested
-    if (!config.caCertificates().isEmpty())
-        q->socket()->setCaCertificates(config.caCertificates());
+    if (!config.caCertificates().isEmpty()) {
+        QSslConfiguration newSslConfig;
+        newSslConfig.setCaCertificates(config.caCertificates());
+        q->socket()->setSslConfiguration(newSslConfig);
+    }
 
     // respect proxy
     q->socket()->setProxy(config.networkProxy());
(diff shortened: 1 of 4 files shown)


src/client/QXmppOutgoingClient.cpp
```diff
@@ -142,7 +143,7 @@ void QXmppOutgoingClientPrivate::connectToHost(const QString &host, quint16 port
 
     // override CA certificates if requested
     if (!config.caCertificates().isEmpty())
-        q->socket()->setCaCertificates(config.caCertificates());
+        QSslConfiguration::defaultConfiguration().setCaCertificates(config.caCertificates());
```
lnjX: What does this change?

lnjX: >  Port deprecated method setCaCertificates of QSslSocket 

If `setCaCertificates` is deprecated, why do you still use it?

lnjX: Nevermind

0xd34df00d: Moreover, doesn't this affect all other sockets in the application using qxmpp? That might be undesirable.

JBBgameich: > If `setCaCertificates` is deprecated, why do you still use it?

This is setCaCertificates of QSslConfiguration not QSslSocket

lnjX: But you shouldn't change the *global* default:
```C++
QSslConfiguration sslConfig;
sslConfig.setCaCertificates(config.caCertificates());
q->socket()->setSslConfiguration(sslConfig);
```

JBBgameich: > But you shouldn't change the _global_ default:
> 
> ```c++
> QSslConfiguration sslConfig;
> sslConfig.setCaCertificates(config.caCertificates());
> q->socket()->setSslConfiguration(sslConfig);
> ```

I noticed that and already pushed a fix

src/client/QXmppOutgoingClient.cpp
```diff
@@ -143,7 +143,7 @@ void QXmppOutgoingClientPrivate::connectToHost(const QString &host, quint16 port
 
     // override CA certificates if requested
     if (!config.caCertificates().isEmpty())
-        QSslConfiguration::defaultConfiguration().setCaCertificates(config.caCertificates());
+        q->socket()->sslConfiguration().setCaCertificates(config.caCertificates());
```
lnjX: AFAIK this won't work, because sslConfiguration() doesn't return a reference.

JBBgameich: Seems you are right there. Unfortunately I have some compilation issues with your suggestion but will push the proper fix once I found it.

JBBgameich:
ping

PR #207, opened 2019-05-17
Port QXmppCallManager to use GStreamer

olesalscheider:
This PR ports QXmppCallManager to use QtGStreamer.
The disadvantage is that it adds QtGStreamer as a dependency. But it also brings the following advantages:
- Nearly 4000 lines less code (-> less maintenance burden)
- We can easily support all codecs supported by GStreamer, also in the future
- The client application has easy access to GStreamer elements, e. g.
 o Widgets for Qt and QML
 o Different audio and video sources including support for screen casting
 o A lot of filters like echo cancelation and noise reduction (cf. webrtcdsp)
- We can easily support secure communication using SRTP (will be added in another patch)

What do you think? Before I continue to clean up this patch I would like to hear your oppinion.


lnjX:
Just a question: Would it be possible to do this with QtMultimedia?

olesalscheider:
QtMultimedia provides device access and encoders / decoders.  But it does not have any of the RTP handling functionality (payloader, depayloader, RTP session management, jitter buffers, encryption, ...) and less filters. I assume we could use QtMultimedia and write these other parts on our own / use some of the code that we already have.

But I don't see the advantages. GStreamer seems to be available on the important platforms (Linux, Windows, OS X, Android, iOS and more) and it has all that is needed for a good RTP experience.
Do you have a reason why you would prefer QtMultimedia?

lnjX:
> Do you have a reason why you would prefer QtMultimedia?

AFAIK most Plasma Mobile/Ubuntu Touch (android) devices don't have hardware acceleration support with GStreamer, but with QtMultimedia. And also just because it's an official part of Qt (less problems with new dependencies).

olesalscheider:
I think we can get hardware acceleration on android by trying the gst-omx codecs before the regular ones. They use the OpenMAX API.

lnjX:
> I think we can get hardware acceleration on android by trying the gst-omx codecs before the regular ones. They use the OpenMAX API.

I'm not talking about android, I'm talking about Halium/libhybris on an android device.
However this is just a nice-to-have and isn't that important (at least for QXmpp).

olesalscheider:
Ah, I see. Well, at least SailfishOS also has hardware acceleration with gst-omx: https://github.com/sailfishos/gst-omx
I'm not sure about the other libhybris based mobile OS though. Sailfish has a few patches for gst-omx and I don't know if they are upstream.

JBBgameich:
QtMultimedia uses gstreamer itself (at least on some platforms), and the accelerated decoding on Plasma Mobile is done as a gstreamer plugin which is then used by QtMultimedia, so using gstreamer here is perfectly fine.

olesalscheider:
The build fails because we don't install qt-gstreamer with homebrew.
I think only gstreamer (but not qt-gstreamer) is available there. But I don't know anything about MacOS - are there other sources for packages? Or would we have to compile it from source?

lnjX:
@olesalscheider The wiki page and README of qt-gstreamer say it's unmaintained and out of date:
https://gstreamer.freedesktop.org/bindings/qt.html
https://github.com/GStreamer/qt-gstreamer/blob/master/README

Is this whole thing a good idea then? How hard would it be to use gstreamer directly or to use the C++ bindings instead?

olesalscheider:
Oh, I was not aware of that. Yes, it seems like using the qt bindings is not a good idea then.
Using gstreamer directly should not be too hard and I would be willing to port the code there if we decide that using gstreamer is a good idea.

lnjX:
@olesalscheider The listed advantages of gstreamer convince me and since it will be an optional dependency (like the codecs before) I see absolutely no problem. :)

olesalscheider:
I ported this to the gstreamer C API. It needs some real-world testing though.
Also now there are a lot of conflicts with the master branch.

src/base/QXmppRtpChannel.h
```diff
@@ -1,298 +0,0 @@
...
-};
-
-/// \brief The QXmppRtpAudioChannel class represents an RTP audio channel to a remote party.
-///
-/// It acts as a QIODevice so that you can read / write audio samples, for
-/// instance using a QAudioOutput and a QAudioInput.
-///
-/// \note THIS API IS NOT FINALIZED YET
-
-class QXMPP_EXPORT QXmppRtpAudioChannel : public QIODevice, public QXmppRtpChannel
```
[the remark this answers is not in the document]
olesalscheider: I think so too.

src/base/QXmppRtcpPacket.h
```diff
@@ -1,169 +0,0 @@
...
-class QXmppRtcpSenderInfo;
-class QXmppRtcpSenderInfoPrivate;
-class QXmppRtcpSourceDescription;
-class QXmppRtcpSourceDescriptionPrivate;
-
-/// \internal
-///
-/// The QXmppRtcpPacket class represents an RTCP packet.
-
-class QXMPP_EXPORT QXmppRtcpPacket
```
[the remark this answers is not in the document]
olesalscheider: Ah, I did not notice that this was already finalized. I can restore the files but they will be unused inside QXmpp. I don't know what would be better...

src/client/QXmppCall_p.h
```diff
@@ -0,0 +1,129 @@
...
+
+    // Media streams
+    QList<QXmppCallStream*> streams;
+    int nextId;
+
+    // Supported codecs
+    QList<GstCodec> videoCodecs = {
+        // vp9enc and x265enc seems to be very slow. Don't offer it for now.
+        //{.pt = 101, .name = "H265", .channels = 1, .clockrate = 90000, .gstPay = "rtph265pay", .gstDepay = "rtph265depay", .gstEnc = "x265enc", .gstDec = "avdec_h265", .encProps = {{"tune", 4}, {"speed-preset", 1}, {"bitrate", 256}}},
+        //{.pt = 100, .name = "VP9", .channels = 1, .clockrate = 90000, .gstPay = "rtpvp9pay", .gstDepay = "rtpvp9depay", .gstEnc = "vp9enc", .gstDec = "vp9dec", .encProps = {{"deadline", 20000}, {"target-bitrate", 256000}}},
```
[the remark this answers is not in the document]
olesalscheider: VP9 is basically unusable, even on very fast computers. But I added them for now with a lower priority.

We can also add the hardware encoders / decoders that are supported by GStreamer to the list. But I think there are none so far for these new codecs.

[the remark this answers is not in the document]
lnjX: Hardware encoders would also be nice for H.264/VP8 - They could be relevant on mobile devices?
(However this isn't that important for now.)

src/base/QXmppRtcpPacket.h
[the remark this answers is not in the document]
lnjX: It probably makes no sense to use the QXmppRtcpPacket without the other not-finalized classes. Then we could just remove this.

@0xd34df00d @jlaine What do you think?

src/client/QXmppCall_p.h
```diff
@@ -0,0 +1,133 @@
...
+    QList<QXmppCallStream*> streams;
+    int nextId;
+
+    // Supported codecs
+    QList<GstCodec> videoCodecs = {
+        {.pt = 100, .name = "H264", .channels = 1, .clockrate = 90000, .gstPay = "rtph264pay", .gstDepay = "rtph264depay", .gstEnc = "x264enc", .gstDec = "avdec_h264", .encProps = {{"tune", 4}, {"speed-preset", 1}, {"bitrate", 256}}},
+        {.pt = 99, .name = "VP8", .channels = 1, .clockrate = 90000, .gstPay = "rtpvp8pay", .gstDepay = "rtpvp8depay", .gstEnc = "vp8enc", .gstDec = "vp8dec", .encProps = {{"deadline", 20000}, {"target-bitrate", 256000}}},
+        // vp9enc and x265enc seem to be very slow. Give them a lower priority for now.
+        {.pt = 102, .name = "H265", .channels = 1, .clockrate = 90000, .gstPay = "rtph265pay", .gstDepay = "rtph265depay", .gstEnc = "x265enc", .gstDec = "avdec_h265", .encProps = {{"tune", 4}, {"speed-preset", 1}, {"bitrate", 256}}},
+        {.pt = 101, .name = "VP9", .channels = 1, .clockrate = 90000, .gstPay = "rtpvp9pay", .gstDepay = "rtpvp9depay", .gstEnc = "vp9enc", .gstDec = "vp9dec", .encProps = {{"deadline", 20000}, {"target-bitrate", 256000}}}
```
lnjX: I'm not sure if I understand the priorities here, but shouldn't the value (`.pt`) here be lower than the values of H.264/VP8 to not prefer them?

Otherwise, this would mean that PCM, AAC and Speex are preferred over Opus which is probably not a good idea.

src/client/QXmppCall.h
```diff
@@ -0,0 +1,123 @@
...
+#include "QXmppCallStream.h"
+
+class QHostAddress;
+class QXmppCallPrivate;
+class QXmppCallManager;
+class QXmppCallManagerPrivate;
+
+/// \brief The QXmppCall class represents a Voice-Over-IP call to a remote party.
+///
+/// \note THIS API IS NOT FINALIZED YET
```
lnjX: Please add `\since QXmpp 1.2` to all new classes or new methods.

olesalscheider: I added it to the new QXmppCallStream class and the new methods in QXmppCall. Is that enough or should I also add it to all new methods in QXmppCallStream?

src/client/QXmppCall_p.h
olesalscheider: The pt value does not matter for the priorities. Only the order in our list / the order of payload-type entries in the containing xmpp message determines the priority.

When it comes to pt values we only have to make sure to follow the standardized values if available and otherwise have a unique 1-1 mapping.

src/client/QXmppCall.h
lnjX: I think it's enough to only add it to the class documentation when the class is new.

lnjX:
There are still some things that aren't tested, many `qFatal` calls (I'm not sure how you can test those?), but also other code. If it's not too hard, I'd be glad if you could add a tests for those stuff. Here's the coverage diff: https://codecov.io/gh/qxmpp-project/qxmpp/pull/207/diff

Apart from that I think we can merge it.

src/client/QXmppCall_p.h
lnjX: Ah, sorry then :)

src/base/QXmppRtcpPacket.h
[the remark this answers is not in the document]
lnjX: Since nobody said anything I'd accept this.

lnjX:
Any idea what's wrong with the windows build?

src/client/QXmppCall.h
```diff
@@ -0,0 +1,123 @@
+/*
+ * Copyright (C) 2008-2019 The QXmpp developers
```
lnjX: Needs to be updated to 2020

src/client/QXmppCallStream.h
```diff
@@ -0,0 +1,71 @@
...
+class QXmppCallStreamPrivate;
+class QXmppIceConnection;
+class QXmppCall;
+class QXmppCallPrivate;
+
+/// \brief The QXmppCallStream class represents an RTP stream in a VoIP call.
+///
+/// \note THIS API IS NOT FINALIZED YET
+///
+/// \since QXmpp 1.2
```
lnjX: Needs to be updated everywhere to QXmpp 1.3

src/client/QXmppCall.cpp
```diff
@@ -0,0 +1,739 @@
...
+        d->manager->callStarted(this);
+
+        // check for call establishment
+        d->setState(QXmppCall::ActiveState);
+    }
+}
+
+/// Returns the GStreamer pipeline.
+///
+/// \since QXmpp 1.2
```
lnjX:
```suggestion
/// \since QXmpp 1.3
```

The \since-tags in this file still need to be updated.

src/base/QXmppStun.cpp
```diff
@@ -2626,7 +2635,19 @@ void QXmppIceConnection::setRemotePassword(const QString &password)
     d->remotePassword = password;
 }
 
-/// Sets the STUN server to use to determine server-reflexive addresses
+/// Sets multiple STUN servers to use to determine server-reflexive addresses
+/// and ports.
+///
+/// \note This may only be called prior to calling bind().
+///
+/// \param servers List of the STUN servers.
```
lnjX:
```suggestion
/// \param servers List of the STUN servers.
///
/// \since QXmpp 1.3
```

src/client/QXmppCallManager.cpp
```diff
@@ -783,16 +152,30 @@ QXmppCall *QXmppCallManager::call(const QString &jid)
     return call;
 }
 
-/// Sets the STUN server to use to determine server-reflexive addresses
+/// Sets multiple STUN servers to use to determine server-reflexive addresses
 /// and ports.
 ///
+/// \note This may only be called prior to calling bind().
+///
+/// \param servers List of the STUN servers.
```
lnjX:
```suggestion
/// \param servers List of the STUN servers.
///
/// \since QXmpp 1.3
```

olesalscheider:
I tested this over the internet with a friend and it worked quite well and quality was good.

More real-world testing with slow hardware or a slow internet connection would be good to see how quality, packet drop and latency are there. Depending on that we can tune the codec parameters a bit and implement the bitrate controller. But maybe that can be done in a second step with some feedback.

lnjX:
Alright, I think we're ready to go then. Do you want me to merge your commits as they are or do you want to squash some of them before?

olesalscheider:
I think it would be good to keep them separate. The first is huge already and the following ones are nicely separated and touch other parts of the code.

JBBgameich:
Hi @olesalscheider,
do you have an example on how to use the new QXmppCallStream API?
I'm struggling with the GstPad callback functions currently. As far as I understand I can only use static functions there, but I also need an object to distinguish between the video and the audio stream. Once I understand how to use them I'd look into writing a QIODevice wrapper again for easier usage with QtMultimedia.

olesalscheider:
You can find a small test program that I wrote here:
https://github.com/olesalscheider/qxmpp/blob/gsttest/voiptest/main.cpp#L11

But I think I understand what your problem is. Should we add an opaque pointer to the callbacks?

JBBgameich:
Thanks for your response. What made the API hard to use for me was that I was trying to use QtMultimedia instead of gstreamer in the client.
Using QAudioInput for this seems more natural for a Qt API, so I think we should ideally have a QAudioInput API for it. Since I now know I can use lambdas for the callbacks (which I tried, but for some reason failed earlier), maybe I can actually implement that myself.

olesalscheider:
Having the GStreamer API also has advantages, though. You can add a lot more elements to the pipeline, e.g. for echo cancellation or other post processing. There are also elements for QML or Qt Widgets to display the video, and elements for screen sharing.

Lambdas work and you can capture the objects you need. The only question is if that is a good design, or if we should add the user object pointer to the API before it is released.

lnjX:
> Having the GStreamer API also has advantages, though. You can add a lot more elements to the pipeline, e.g. for echo cancellation or other post processing. There are also elements for QML or Qt Widgets to display the video, and elements for screen sharing.

Maybe it's possible to provide access to the gstreamer pipeline, but still to wrap the output of it into a QAudioOutput and to accept a QAudioInput. If using the gstreamer APIs works fine, we can maybe also just keep that, though.

For the QML elements you probably refer to QtGstreamer, but since it's unmaintained I don't think we should recommend using that.

> You can find a small test program that I wrote here:
> https://github.com/olesalscheider/qxmpp/blob/gsttest/voiptest/main.cpp#L11

I think we should also add your example to QXmpp. I think that's very helpful as there's no detailed description in the documentation on how to use it.

A note about STUN:
Some XMPP servers already provide STUN servers - Do you know whether we can automatically recognize and use them?

olesalscheider:
No, there are also Qt elements (e.g. qmlglsink) in gst-plugins-good:
https://github.com/GStreamer/gst-plugins-good/tree/master/ext/qt

But so far I haven't used them, though.

We can add my example but it needs to be cleaned up a bit, I guess.

About STUN: I think XEP-0215 is not implemented. How common is it for servers to support this?

lnjX:
> No, there are also Qt elements (e.g. qmlglsink) in gst-plugins-good:
> https://github.com/GStreamer/gst-plugins-good/tree/master/ext/qt

Ah, thanks, that's good to know.


> About STUN: I think XEP-0215 is not implemented. How common is it for servers to support this?

There's a module for prosody, but it's probably not used by many servers. I know that ejabberd has a STUN server enabled by default, but ejabberd (at least the open-source edition) is not supporting XEP-0215. There's also https://wiki.xmpp.org/web/SRV_Records which recommends setting up SRV records for STUN/TURN. Maybe this is also an option.

PR #210, opened 2019-09-04
Replace Q_FOREACH (foreach) by C++11 ranged for-loops

lnjX:
Q_FOREACH is bad and will be deprecated in the future:
https://www.kdab.com/goodbye-q_foreach/

This also disables Q_FOREACH in the future by defining QT_NO_FOREACH.


src/base/QXmppStun.cpp
```diff
@@ -1248,7 +1248,7 @@ void QXmppTurnAllocation::disconnectFromHost()
 
     // clear channels and any outstanding transactions
     m_channels.clear();
-    foreach (QXmppStunTransaction *transaction, m_transactions)
+    for (auto *transaction : m_transactions)
```
tehnick: Just interesting: why `qDeleteAll()` is not used here?

tehnick: Ok, as I see:
```
$ grep -r qDeleteAll * --exclude-dir=builddir/ | wc -l
0
```
And this is the answer.

tehnick:
+1 for this change.

tehnick:
`qAsConst()` function was introduced in Qt 5.7, but in Ubuntu Xenial is Qt 5.5.1.

tehnick:
Are you going to drop Ubuntu Xenial support or these changes will be fixed?

lnjX:
> Are you going to drop Ubuntu Xenial support or these changes will be fixed?

I think Qt 5.4 is so old that nobody actually develops on it. We could introduce a policy to support the last two LTS versions of Qt, I think that would be enough.

About travis-ci: I think we can just update to bionic.

src/base/QXmppStun.cpp
lnjX: That's definitely the better way.

tehnick:
> I think Qt 5.4 is so old that nobody actually develops on it.

Why are you talking about Qt 5.4? In Ubuntu Xenial Qt 5.5.x is used.
In our official PPA I still update packages for it:
https://launchpad.net/~tehnick/+archive/ubuntu/qxmpp-devel/+packages
https://launchpad.net/~tehnick/+archive/ubuntu/qxmpp/+packages
(This is just for a note. I may disable updates for Ubuntu Xenial in any moment.)

> We could introduce a policy to always support the last two LTS versions of Qt (currently that would be Qt 5.9), I think that would be enough.

Personally I would prefer to save support of Qt 5.6.x in the nearest release of qxmpp. Though Qt 5.7.x is also not worst variant (it is still suitable for WinXP), but Qt 5.6.x is better because there were a number of bug fix releases while it has been supported as LTS release. (Yep, there are some programs which still support WinXP...)

As for the future releases it would be nice to be oriented on Qt versions in Debian stable and latest Ubuntu LTS at least, because I backport qxmpp packages for them. As you know they usually do not use Qt LTS releases...

tehnick:
> As for the future releases it would be nice to be oriented on Qt versions in Debian stable and latest Ubuntu LTS at least, because I backport qxmpp packages for them. As you know they usually do not use Qt LTS releases...

For example:
* Ubuntu Bionic (18.04) uses Qt 5.9.5 + patches
* Debian Buster (10) uses Qt 5.11.3 + patches
* Ubuntu ??? 20.04 LTS will be released in April 2020
* Ubuntu Bionic (18.04) will be supported until April 2023, but it would be enough to support qxmpp build until April 2021 (most of active users will switch to Ubuntu 20.04 from it)
* Debian Buster (10) will be supported until 2022

Qt 5.9.x LTS will be supported until May 2020 and this is not enough for supporting of Ubuntu Bionic as you see...

src/base/QXmppRtpChannel.cpp
```diff
@@ -490,9 +490,7 @@ qint64 QXmppRtpAudioChannel::readData(char * data, qint64 maxSize)
 void QXmppRtpAudioChannel::payloadTypesChanged()
 {
     // delete incoming codecs
-    foreach (QXmppCodec *codec, d->incomingCodecs)
-        delete codec;
-    d->incomingCodecs.clear();
+    qDeleteAll(d->incomingCodecs);
```
tehnick: Hey! You forgot to call `clear()` here. Please re-read `qDeleteAll()` usage in documentation.

src/base/QXmppRtpChannel.cpp
```diff
@@ -921,10 +918,9 @@ QIODevice::OpenMode QXmppRtpVideoChannel::openMode() const
 void QXmppRtpVideoChannel::payloadTypesChanged()
 {
     // refresh decoders
-    foreach (QXmppVideoDecoder *decoder, d->decoders)
-        delete decoder;
-    d->decoders.clear();
-    foreach (const QXmppJinglePayloadType &payload, m_incomingPayloadTypes) {
+    qDeleteAll(d->decoders);
```
tehnick: One more missed `clear()`.

src/base/QXmppStun.cpp
```diff
@@ -1873,16 +1873,14 @@ void QXmppIceComponentPrivate::setSockets(QList<QUdpSocket*> sockets)
 
     // clear previous candidates and sockets
     localCandidates.clear();
-    foreach (CandidatePair *pair, pairs)
-        delete pair;
-    pairs.clear();
-    foreach (QXmppIceTransport *transport, transports)
+    qDeleteAll(pairs);
```
tehnick: One more missed `clear()`.

src/base/QXmppDataForm.cpp
```diff
@@ -637,8 +637,7 @@ void QXmppDataForm::toXml(QXmlStreamWriter *writer) const
             type == QXmppDataForm::Field::ListSingleField)
         {
             QPair<QString, QString> option;
-            foreach (option, field.options())
-            {
+            for (const auto &option : field.options()) {
```
0xd34df00d: Note this shadows the other `option`. I guess the previous line can (and should) be removed.

src/base/QXmppDiscoveryIq.cpp
```diff
@@ -220,7 +220,7 @@ QByteArray QXmppDiscoveryIq::verificationString() const
 
             QStringList keys = fieldMap.keys();
             std::sort(keys.begin(), keys.end());
```
0xd34df00d: While we're at this, I'm not sure `keys` need to be sorted. The joy of tree-based `QMap` is that it naturally stores them in order.

src/base/QXmppJingleIq.cpp
```diff
@@ -590,7 +590,7 @@ QString QXmppJingleIq::Content::toSdp() const
     // media
     QString payloads;
     QStringList attrs;
-    foreach (const QXmppJinglePayloadType &payload, d->payloadTypes) {
+    for (const QXmppJinglePayloadType &payload : d->payloadTypes) {
```
0xd34df00d: Maybe make this `auto` too?

src/base/QXmppRtpChannel.cpp
```diff
@@ -352,7 +352,7 @@ void QXmppRtpAudioChannel::datagramReceived(const QByteArray &ba)
     QXmppCodec *codec = 0;
     const quint8 packetType = packet.type();
     if (!d->incomingCodecs.contains(packetType)) {
-        foreach (const QXmppJinglePayloadType &payload, m_incomingPayloadTypes) {
+        for (const auto &payload : m_incomingPayloadTypes) {
```
0xd34df00d: I'd much rather rewrite this with `std::find_if` to better express the intent, but ultimately it's up to you.

src/base/QXmppRtpChannel.cpp
0xd34df00d: Quite frankly I feel like this code is just asking for `std::unique_ptr`s. Manual memory management, explicit deletes, eww... But this may be a task for another day and another PR.

src/base/QXmppSasl.cpp
```diff
@@ -923,18 +923,15 @@ QMap<QByteArray, QByteArray> QXmppSaslDigestMd5::parseMessage(const QByteArray &
...
             ba.append(',');
         ba.append(key + "=");
         QByteArray value = map[key];
         const char *separators = "()<>@,;:\\\"/[]?={} \t";
         bool quote = false;
-        for (const char *c = separators; *c; c++)
-        {
-            if (value.contains(*c))
-            {
+        for (const char *c = separators; *c; c++) {
```
0xd34df00d: Prefer `std::any_of` (or, perhaps even better, `std::find_first_of`).

src/base/QXmppSasl.cpp
```diff
@@ -923,18 +923,15 @@ QMap<QByteArray, QByteArray> QXmppSaslDigestMd5::parseMessage(const QByteArray &
 QByteArray QXmppSaslDigestMd5::serializeMessage(const QMap<QByteArray, QByteArray> &map)
 {
     QByteArray ba;
-    foreach (const QByteArray &key, map.keys())
-    {
+    for (const auto &key : map.keys()) {
```
0xd34df00d: This allocates a whole list of keys and then does a lookup for each key a few lines later. Perhaps it'd be better to do old-style iterator-based `for` loop (or, alternatively, I have a handy view class for exactly this case, but it requires something between C++14 and 17).

src/client/QXmppCallManager.cpp
```diff
@@ -117,15 +117,15 @@ QXmppCallPrivate::QXmppCallPrivate(QXmppCall *qq)
 
 QXmppCallPrivate::Stream *QXmppCallPrivate::findStreamByMedia(const QString &media)
 {
-    foreach (Stream *stream, streams)
+    for (auto *stream : streams)
```
0xd34df00d: Prefer `std::find_if`.

src/client/QXmppCallManager.cpp
```diff
@@ -117,15 +117,15 @@ QXmppCallPrivate::QXmppCallPrivate(QXmppCall *qq)
...
         if (stream->media == media)
             return stream;
     return 0;
 }
 
 QXmppCallPrivate::Stream *QXmppCallPrivate::findStreamByName(const QString &name)
 {
-    foreach (Stream *stream, streams)
+    for (auto *stream : streams)
```
0xd34df00d: ditto (and in a bunch of other places that follow in this file). Let me know if you'd consider this something you'd like to change, and I'll mark the appropriate locations.

src/base/QXmppSasl.cpp
lnjX: What about `std::for_each`?

src/base/QXmppRtpChannel.cpp
lnjX: Sorry I should have read the documentation of `qDeleteAll()` first.

src/client/QXmppCallManager.cpp
lnjX: I think I got most of them, but feel free to check again :)

lnjX:
> > I think Qt 5.4 is so old that nobody actually develops on it.
> 
> Why are you talking about Qt 5.4? In Ubuntu Xenial Qt 5.5.x is used.
> In our official PPA I still update packages for it:
> https://launchpad.net/~tehnick/+archive/ubuntu/qxmpp-devel/+packages
> https://launchpad.net/~tehnick/+archive/ubuntu/qxmpp/+packages
> (This is just for a note. I may disable updates for Ubuntu Xenial in any moment.)

Sorry, I wanted to say Qt 5.5. Xenial support is nice as long as it does not make things too complicated.

> > We could introduce a policy to always support the last two LTS versions of Qt (currently that would be Qt 5.9), I think that would be enough.
> 
> Personally I would prefer to save support of Qt 5.6.x in the nearest release of qxmpp. Though Qt 5.7.x is also not worst variant (it is still suitable for WinXP), but Qt 5.6.x is better because there were a number of bug fix releases while it has been supported as LTS release. (Yep, there are some programs which still support WinXP...)

I would need to add many preprocessor checks to support Qt 5.6. It's the question whether it is worth it - we will sooner or later require Qt > 5.6 anyway I think.

> As for the future releases it would be nice to be oriented on Qt versions in Debian stable and latest Ubuntu LTS at least, because I backport qxmpp packages for them. As you know they usually do not use Qt LTS releases... (Ubuntu Bionic is an accidental exception.)

I am totally fine with that. :)

src/base/QXmppDiscoveryIq.cpp
lnjX: I dont want to touch the discoveryiq too much in this pr.

lnjX:
@tehnick So let's do a QXmpp 1.1 release (with full Qt 5.0+ support) soon and merge this after that. QXmpp 1.2 would then require Qt 5.7+.

tehnick:
> So let's do a QXmpp 1.1 release (with full Qt 5.0+ support) soon and merge this after that. QXmpp 1.2 would then require Qt 5.7+.

Ok. Go ahead. I'll try to update Debian packages during this week.

tehnick:
@lnjX I was busy at work during last months and had not looked at project activity. Sorry, if I have missed other highlights.

lnjX:
@tehnick No problem at all, also not much has happened in the meantime. :)

PR #214, opened 2019-09-27
Deprecate QXmppClient::{rosterManager,vCardManager,versionManager}()

lnjX:


jlaine:
I think I agree with this, these hard-coded manager accessors don't really follow the modular philosophy QXmpp eventually embraced.

I do think we need a better commit message + entry in the README explaining the rationale for this change with a clear deprecation timeline.

Anyone else have thoughts on this?

lnjX:
> I do think we need a better commit message + entry in the README explaining the rationale for this change with a clear deprecation timeline.

For the deprecation timeline, I think we can just keep those methods with this deprecation warning until QXmpp 2.0 / a major release (breaking source compatibility).

I don't see the need for a README entry? I mean there's already a compiler warning, so if anyone uses those methods they will be notified about it.

PR #215, opened 2019-10-14
Use QSharedData for all private data classes

lnjX:
(Was my fault in most/all cases)

### DIFF
--- a/src/base/QXmppHttpUploadIq.h
+++ b/src/base/QXmppHttpUploadIq.h
@@ -24,11 +24,12 @@
 #ifndef QXMPPHTTPUPLOADIQ_H
 #define QXMPPHTTPUPLOADIQ_H
 
-#include <QMap>
-#include <QMimeType>
-#include <QUrl>
+#include <QSharedDataPointer>
 
 #include "QXmppIq.h"
+
+class QUrl;
+class QMimeType;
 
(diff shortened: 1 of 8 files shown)


src/base/QXmppHttpUploadIq.h
```diff
@@ -44,6 +45,8 @@ class QXMPP_EXPORT QXmppHttpUploadRequestIq : public QXmppIq
     QXmppHttpUploadRequestIq();
     ~QXmppHttpUploadRequestIq() override;
 
+    QXmppHttpUploadRequestIq &operator=(const QXmppHttpUploadRequestIq &);
```
0xd34df00d: Adding move ctors and move assignment operators might be a good idea for this and similar classes

lnjX: Right. Just what I'm not sure about are QObject-based classes. Does it even make sense to use QSharedData there? - QObject-based classes cannot be copied usually.

lnjX: I added missing move ctors to the MIX item classes and HTTP Upload IQs.

I think I should better keep QXmppStreamPrivate and the RTP channel classes as they are - the only advantage of QSharedData here is that you don't need to manually delete the objects in the end.

PR #224, opened 2019-10-20
Modernize cpp

JBBgameich:
by using raw string literals, range based loops were possible and auto.

### DIFF
--- a/src/base/QXmppStun.cpp
+++ b/src/base/QXmppStun.cpp
@@ -1439,7 +1439,7 @@ void QXmppTurnAllocation::setState(AllocationState state)
 
 void QXmppTurnAllocation::transactionFinished()
 {
-    QXmppStunTransaction *transaction = qobject_cast<QXmppStunTransaction*>(sender());
+    auto *transaction = qobject_cast<QXmppStunTransaction*>(sender());
     if (!transaction || !m_transactions.removeAll(transaction))
         return;
     transaction->deleteLater();
@@ -1825,7 +1825,7 @@ bool QXmppIceComponentPrivate::addRemoteCandidate(const QXmppJingleCandidate &ca
         if (!isCompatibleAddress(local.host(), candidate.host()))
             continue;
 
-        CandidatePair *pair = new CandidatePair(component, config->iceControlling, q);
+        auto *pair = new CandidatePair(component, config->iceControlling, q);
         pair->remote = candidate;
         pair->transport = transport;
         pairs << pair;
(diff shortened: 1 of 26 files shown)


src/base/QXmppStun.cpp
```diff
@@ -2350,11 +2350,11 @@ static QList<QUdpSocket*> reservePort(const QList<QHostAddress> &addresses, quin
...
     QList<QUdpSocket*> sockets;
     foreach (const QHostAddress &address, addresses) {
-        QUdpSocket *socket = new QUdpSocket(parent);
+        auto *socket = new QUdpSocket(parent);
         sockets << socket;
         if (!socket->bind(address, port)) {
-            for (int i = 0; i < sockets.size(); ++i)
-                delete sockets[i];
+            for (auto &socket : sockets)
+                delete socket;
```
lnjX: `qDeleteAll(sockets);`

src/base/QXmppStun.cpp
```diff
@@ -2439,8 +2439,8 @@ QList<QUdpSocket*> QXmppIceComponent::reservePorts(const QList<QHostAddress> &ad
 
         // cleanup if we failed
         if (sockets.size() != expectedSize) {
-            for (int i = 0; i < sockets.size(); ++i)
-                delete sockets[i];
+            for (auto &socket : sockets)
+                delete socket;
```
lnjX: `qDeleteAll(sockets);`

src/client/QXmppClient.h
```diff
@@ -128,9 +128,9 @@ class QXMPP_EXPORT QXmppClient : public QXmppLoggable
     T* findExtension()
     {
         QList<QXmppClientExtension*> list = extensions();
-        for (int i = 0; i < list.size(); ++i)
+        for (auto i : list)
```
lnjX: Make the container const please

src/base/QXmppStun.cpp
0xd34df00d: Personally I'd prefer using `std::unique_ptr` or `std::shared_ptr` for this.

src/base/QXmppStun.cpp
0xd34df00d: ditto

src/client/QXmppClient.h
```diff
@@ -127,10 +127,10 @@ class QXMPP_EXPORT QXmppClient : public QXmppLoggable
     template<typename T>
     T* findExtension()
     {
-        QList<QXmppClientExtension*> list = extensions();
-        for (int i = 0; i < list.size(); ++i)
+        const QList<QXmppClientExtension*> list = extensions();
+        for (auto i : list)
```
lnjX: Sorry, but could you also rename `i` to `ext` or `extension`?

src/base/QXmppStun.cpp
lnjX: We actually can't use `std::unique_ptr`s, because then we would need to change the return type and so the public API. `reservePort()` is used in here and it wouldn't make sense to change its return type.

src/base/QXmppStun.cpp
lnjX: See below

lnjX: `qDeleteAll();`

PR #227, opened 2019-10-26
Replace deprecated Q_ENUMS with Q_ENUM

lnjX:
Q_ENUM exists since Qt 5.5, more details can be found here:
https://woboq.com/blog/q_enum.html


lnjX:
@0xd34df00d Yes, a clang-format file would be great... QXmpp has a different formatting everywhere...

PR #232, opened 2019-11-28
Fix CSI stanzas are sent when not connected, Fix CSI state is not reset on connect

lnjX:


lnjX:
@0xd34df00d Thanks! :)

PR #238, opened 2019-12-18
QXmppGlobal: Use QT_VERSION_CHECK() to generate QXMPP_VERSION

lnjX:
QT_VERSION_CHECK() returns the same format as is currently used, it is
easier to read and fixes a problem that could occur with the previous
marco. The macro was unenclosed causing comparison to fail. This also
happened in Kaidan [1].

[1]: https://invent.kde.org/kde/kaidan/commit/5a3e5e8a748488f4d55302b4beacce880dde381a

### DIFF
--- a/src/base/QXmppGlobal.h.in
+++ b/src/base/QXmppGlobal.h.in
@@ -46,7 +46,7 @@
 /// available.
 ///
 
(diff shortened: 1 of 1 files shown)


lnjX:
@0xd34df00d Thanks! :)

PR #239, opened 2019-12-23
Add a .clang-format somewhat close to the current style

0xd34df00d:
Did somebody ask for a `.clang-format`? :)

I took a few files that seem to be representative of the general project style (namely, `QXmppMessage.*` and `QXmppClient.*`) and fed them to a little program I've been working on lately that tries to find the closest matching style. Here's the result which I expect to be more or less close to the best achievable with `clang-format`.

Turns out, that's still quite a significant change, so that's up for you folks to consider if we want to go this route. Personally, the thing I hate the most is the extra space before `&` when there is no parameter name, like `foo(const Bar &)`, but clang-format doesn't have a handle to control that, and aligning all pointers/references to the left makes this change twice as big.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,15 @@
+AlignAfterOpenBracket: Align
+AlignTrailingComments: true
+BasedOnStyle: WebKit
+BraceWrapping:
+  AfterClass: true
+  AfterFunction: true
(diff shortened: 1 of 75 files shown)


JBBgameich:
Thanks, this will help me with the connects pull request :)

PR #244, opened 2020-01-29
Replace manual xmlns wrting by writeDefaultNamespace()

lnjX:
This is the result of:

```bash
sed -i 's/writeAttribute("xmlns", /writeDefaultNamespace(/g' $(find . -iname "*.cpp" -or -iname "*.h")
```


lnjX:
@0xd34df00d Thanks for the approval :)

PR #245, opened 2020-01-31
QXmppClientExtension: Add preferredInsertionIndex()

lnjX:
This is very useful for extensions with dependencies.

### DIFF
--- a/src/client/QXmppClient.cpp
+++ b/src/client/QXmppClient.cpp
@@ -143,11 +143,19 @@ QXmppClient::~QXmppClient()
 
 /// Registers a new \a extension with the client.
 ///
+/// The new extension is inserted at its preferred index
+/// (QXmppClientExtension::preferredInsertionIndex()). If the extension has no
+/// preference on this, it is appended (regular case).
+///
 /// \param extension
 
-bool QXmppClient::addExtension(QXmppClientExtension* extension)
+bool QXmppClient::addExtension(QXmppClientExtension *extension)
 {
-    return insertExtension(d->extensions.size(), extension);
+    int index = extension->preferredInsertionIndex(this);
+    if (index < 0)
+        index = d->extensions.size();
+
+    return insertExtension(index, extension);
 }
 
 /// Registers a new \a extension with the client at the given \a index.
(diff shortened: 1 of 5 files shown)


lnjX:
I'm not sure I got you right. Are you suggesting to add something like the following?
 * `QList<QXmppClientExtension> QXmppClientExtension::extensionsToLoadAfter()`
 * `QList<QXmppClientExtension> QXmppClientExtension::extensionsToLoadBefore()`

0xd34df00d:
Yeah, what I had in mind is somewhat close to this, although it operates on type level as opposed to value level — if the `extensionsToLoadAfter()`/`Before()` functions return a list of `QXmppClientExtension`s you'd have to return actual values in those lists, and that's probably overly complicating things.

What I had in mind is a function like
```cpp
template<typename... Args>
struct Before {};

template<typename... Args>
struct After {};

template<typename... BeforeExts, typename... AfterExts>
QXmppClient::addExtension(QXmppClientExtension *ext, Before<BeforeExts>, After<AfterExts>) {
    // determine the indices of extensions with types mentioned in BeforeExts and AfterExts
    // ignoring the ones that don't exist yet, and calculate the corresponding insertion index
    // in the obvious way
}
```
maybe with some additional overloads in case there are no `Before` or no `After` constraints.

I probably omitted something important, so please let me know if you'd like me to elaborate on smth!

lnjX:
That approach would require the user to use the correct Before<> and After<> arguments. I wounder whether you could add two virtual getters to the QXmppClientExtension instead...

Also I'm not sure how to deal with the parameters. -- I mean I can't iterate on them using a for-loop of course. I only found examples with recursive functions, processing one template parameter at a time.

I actually found out that I don't need this anymore, so it's currently not a priority for me. (still it would be a nice feature, although I don't know whether anyone really needs it)

0xd34df00d:
Yeah, I thought about that after sending the previous comment. You probably could overcome that with more template magic, but since you don't need this specific feature anymore, I probably won't get into too much details about that.

And, since it's not a priority for you then I could take a look at implementing this later on when I get a chance to look at it. Sounds like a fun and useful thing to do!

PR #258, opened 2020-02-11
Make QXmpp work with projects using QT_NO_KEYWORDS

lnjX:
This replaces all occurencies of 'slots' and 'signals' with 'Q_SLOTS'
and 'Q_SIGNALS'. This allows for smooth integration with software
projects that need QT_NO_KEYWORDS, such as those ones that rely on
boost libraries.

Closes #115.

Co-authored-by: Tommaso Cucinotta <tommaso.cucinotta@santannapisa.it>

Replaces #115.


lnjX:
Thanks for the review. :)

PR #259, opened 2020-02-11
ci: appveyor: Try to enable parallel builds

lnjX:


lnjX:
Ok, at least the 2017 build is a bit faster ...

PR #260, opened 2020-02-24
Fix switch statement when handling IQ stanza for registration

melvo:


lnjX:
Thank you.

Now we also need to improve the unit tests to prevent that something like this happens again. Could you do that?

lnjX:
Could you change the base branch to `release/1.2.1`, so we can make a patch release with this?

melvo:
The base branch is changed and the test added.

tests/qxmppregistrationmanager/tst_qxmppregistrationmanager.cpp
```diff
@@ -370,7 +370,14 @@ void tst_QXmppRegistrationManager::testRegistrationResult()
...
     manager->handleStanza(writePacketToDom(serverResult));
 
-    QVERIFY(signalCalled);
+    if (isSuccess) {
+        QVERIFY(succeededSignalCalled);
+        QVERIFY(!failedSignalCalled);
+    } else {
+        QVERIFY(!succeededSignalCalled);
+        QVERIFY(failedSignalCalled);
+    }
```
lnjX: I think this is even better:
```c++
    QCOMPARE(succeededSignalCalled, isSuccess);
    QCOMPARE(failedSignalCalled, !isSuccess);
```

PR #262, opened 2020-03-19
Do not include 'ask' attribute when renaming roster item

melvo:
The *ask* attribute must not be included. See https://tools.ietf.org/html/rfc6121#section-2.1.2.2.

### DIFF
--- a/src/client/QXmppRosterManager.cpp
+++ b/src/client/QXmppRosterManager.cpp
@@ -282,6 +282,10 @@ bool QXmppRosterManager::renameItem(const QString &bareJid, const QString &name)
     QXmppRosterIq::Item item = d->entries.value(bareJid);
     item.setName(name);
 
+    // If there is a pending subscription, do not include the corresponding attribute in the stanza.
+    if (!item.subscriptionStatus().isEmpty())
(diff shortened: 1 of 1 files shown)


src/client/QXmppRosterManager.cpp
```diff
@@ -282,6 +282,10 @@ bool QXmppRosterManager::renameItem(const QString &bareJid, const QString &name)
     QXmppRosterIq::Item item = d->entries.value(bareJid);
     item.setName(name);
 
+    // If there is a pending subscription, do not include the corresponding attribute in the stanza.
+    if (!item.subscriptionStatus().isEmpty())
+        item.setSubscriptionStatus(QString{});
```
lnjX: Better use `QString()` or `{}` instead of `QString{}`.

lnjX:
This should probably also get into QXmpp 1.2.1, so could you change the base branch to `release/1.2.1`?

Also, ideally, we need a unit test for this. Unfortunately there are no unit tests for the registration manager at all, but It would be nice if you could add tests for at least the rename method. You can have a look at the tests of the registration manager for some examples on how to test incoming/outgoing xml.

PR #263, opened 2020-03-22
base: Convert to QStringLiteral

JBBgameich:
Added two short forms as defines:
```
QStringLiteral -> QSL
QByteArrayLiteral -> QBL
```

This was necessary to make this amount of work bearable, but it should be easy to expand them if you like that better.

### DIFF
--- a/src/base/QXmppSasl.cpp
+++ b/src/base/QXmppSasl.cpp
@@ -27,6 +27,7 @@
 
 #include <cstdlib>
 
+#include <QByteArray>
 #include <QDomElement>
 #include <QMessageAuthenticationCode>
 #include <QStringList>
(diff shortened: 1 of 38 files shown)


src/base/QXmppSasl.cpp
```diff
@@ -924,9 +925,9 @@ QByteArray QXmppSaslDigestMd5::serializeMessage(const QMap<QByteArray, QByteArra
             }
         }
         if (quote) {
-            value.replace("\\", "\\\\");
-            value.replace("\"", "\\\"");
-            ba.append("\"" + value + "\"");
+            value.replace(BL("\\"), BL("\\\\"));
```
JBBgameich: This might need some more testing, not sure how those escape characters end up after a QStringLiterals

lnjX: This shouldn't be a problem. Before the characters were converted exactly the same way, but on runtime.

JBBgameich:
- [x] I'd prefer "QSL" and "QBL", I think you can more easily understand the abbrieviation.
- [x] There're already "QStringLiteral"s used in the code, could you replace them by the short ones, too?

PR #264, opened 2020-03-23
Set user's vCard also when 'from' attribute contains user's bare JID

melvo:


src/client/QXmppVCardManager.cpp
```diff
@@ -113,7 +113,8 @@ bool QXmppVCardManager::handleStanza(const QDomElement& element)
         QXmppVCardIq vCardIq;
         vCardIq.parse(element);
 
-        if (vCardIq.from().isEmpty()) {
+        QString from = vCardIq.from();
+        if (from.isEmpty() || from == client()->configuration().jidBare()) {
```
lnjX:
```suggestion
        if (vCardIq.from().isEmpty() || vCardIq.from() == client()->configuration().jidBare()) {
```

This way you don't need to copy the QString.

src/client/QXmppVCardManager.cpp
```diff
@@ -113,14 +113,29 @@ bool QXmppVCardManager::handleStanza(const QDomElement& element)
...
+                emit noVCardFound();
+                return true;
+                break;
+            case QXmppIq::Error::ItemNotFound:
+                emit noClientVCardFound();
+                return true;
+                break;
+            default:
+                return false;
+            }
```
lnjX: The XEP says under [section 3.1](https://xmpp.org/extensions/xep-0054.html#sect-idm46269769635936):
> If no vCard exists, the server MUST return a stanza error (which SHOULD be &lt;item-not-found/&gt;) _**OR**_ an IQ-result containing an empty <vCard/> element.

So this means the error is optional and the signals won't work sometimes. Also I don't think we really need the error signals since you can always check whether the returned VCard is empty or the returned VCardIq contains an error.

This would break compatibility with users relying on receiving error IQs via the normal signal. And adding new signals would mean that we can't include it into 1.2.1, because it's not a 'fix' then. (https://semver.org/)

src/client/QXmppVCardManager.cpp
```diff
@@ -113,14 +113,29 @@ bool QXmppVCardManager::handleStanza(const QDomElement& element)
...
+        }
+
+        if (vCardIq.from().isEmpty() || vCardIq.from() == client()->configuration().jidBare()) {
             d->clientVCard = vCardIq;
             d->isClientVCardReceived = true;
             emit clientVCardReceived();
+            return true;
```
lnjX: This breaks compatibility.

lnjX: It changes the behaviour of the vCardReceived() signal. Clients depending on that the signal is always emitted wouldn't work after this anymore.

lnjX:
Replaced by #281.

PR #266, opened 2020-03-24
[tests] run tests on Windows

jlaine:


jlaine:
@lnjX this is pretty infuriating, I'm able to build the code and tests under Windows, however running the tests fails with a "DLL not found" error. My guess is that the QXmpp DLL is not being found as it's not on the PATH.

lnjX:
I'm sorry, but I don't think I can help you with the windows-specific things. You're in the right directory, right? -- I've no idea then.

jlaine:
I haven't used Windows for the past 20 years so I'm not an ideal candidate for this either :)

I think the problem is that the DLL gets built in one directory (probably under src/ somewhere) whereas the test suite binaries are somewhere in tests/. Linux and OS X have a concept of RPATH, Windows does not: the DLL files need to be on the PATH or in the same directory as the binary.

I read that using `set_tests_properties` is an option to modify the PATH when executing the test suite, but my CMake skills are rusty and I don't know how to retrieve the directory in which the `qxmpp` library got built:

https://stackoverflow.com/questions/54560434/shared-library-under-windows-and-cmake-dll-not-found-before-installation

olesalscheider:
I think you can get the path using $<TARGET_FILE_DIR:qxmpp>. If that path is added to the PATH variable then the tests should run.

lnjX (approved):

PR #267, opened 2020-03-25
 Use 'QRandomGenerator::global()->generate()' instead of deprecated 'qsrand()' and 'qrand()' for Qt >= 5.10

melvo:


0xd34df00d:
Note that this is Qt ≥5.10-only feature. I'd probably postpone this change until at least end of life of some existing supported linux distros releases that have Qt 5.9 or earlier.

lnjX:
Yes I agree with @0xd34df00d on this. What you could do though is to only use QRandomGenerator when compiling against Qt >= 5.10.

melvo:
@lnjX QRandomGenerator is used for Qt >= 5.10 now.

src/base/QXmppUtils.cpp
```diff
@@ -280,8 +283,11 @@ int QXmppUtils::generateRandomInteger(int N)
 {
     Q_ASSERT(N > 0 && N <= RAND_MAX);
     int val;
-    while (N <= (val = qrand() / (RAND_MAX / N))) {
-    };
+#if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
+    while (N <= (val = QRandomGenerator::global()->generate() / (RAND_MAX / N))) { }
+#else
+    while (N <= (val = qrand() / (RAND_MAX / N))) { }
```
lnjX: Can you run clang-format on this? (`clang-format -i src/base/QXmppUtils.cpp` from repository root)

PR #277, opened 2020-03-30
Deprecate PaymentRequired error condition

lnjX:
The <payment-required/> error condition was not adopted in RFC6120 as it
was unused.


0xd34df00d:
Thanks for cleaning up the code base.

PR #282, opened 2020-04-01
QXmppCallStream: Wrap callback in std::function

JBBgameich:
Allows to use non-static functions (or lambdas with captures)

This should even to some extend be source compatible.

### DIFF
--- a/src/client/QXmppCallStream.h
+++ b/src/client/QXmppCallStream.h
@@ -28,6 +28,8 @@
 
 #include <gst/gst.h>
 
+#include <functional>
+
 #include <QObject>
 
 class QXmppCallStreamPrivate;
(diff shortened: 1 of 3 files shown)


src/client/QXmppCallStream.h
```diff
@@ -34,6 +34,8 @@ class QXmppCallStreamPrivate;
 class QXmppIceConnection;
 class QXmppCall;
 class QXmppCallPrivate;
+template<typename _Signature>
+    class function;
```
0xd34df00d, line 40: STL classes shouldn't be forward-declared generally (as in, the standard explicitly forbids this), and this decl is a no-op anyway since it's not in `std::`.

JBBgameich: Thanks, its fixed now. Apparently the Qt 5.15 headers are leaking the functional header themselves, so it still compiled for me.

olesalscheider:
This looks good to me.

Yes, it breaks ABI, but this was not in any released version anyway...

PR #294, opened 2020-04-19
Skip stream errors on successful account deletion

melvo:
Some servers send errors after successfully deleting an account. On https://xmpp.org/extensions/xep-0077.html#usecases-cancel, there is only the stream error `<not-authorized/>` specified but during testing I found two error types sent by the tested servers.

### DIFF
--- a/src/client/QXmppClient.h
+++ b/src/client/QXmppClient.h
@@ -3,6 +3,7 @@
  *
  * Author:
  *  Manjeet Dahiya
+ *  Melvin Keskin
  *
  * Source:
  *  https://github.com/qxmpp-project/qxmpp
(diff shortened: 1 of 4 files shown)


src/client/QXmppClient.h
```diff
@@ -180,6 +181,8 @@ class QXMPP_EXPORT QXmppClient : public QXmppLoggable
     bool isAuthenticated() const;
     bool isConnected() const;
 
+    void setIsAccountBeingDeleted(bool isAccountBeingDeleted);
```
lnjX, line 184: This is no nice way of handling that. A nice way would be something like a hook that is called that managers could handle.

But in general I don't think that either the user (using the Client) or managers should have to handle or should be able to handle stream errors.

Issue #297, opened 2020-05-09
Linking against static QXmpp


leobasilio:
When building a static version of QXmpp using Visual Studio, linking against it fails because of the following on QXmppGlobal.h:

```cpp
#if defined(QXMPP_BUILD)
#    define QXMPP_EXPORT Q_DECL_EXPORT
#else
#    define QXMPP_EXPORT Q_DECL_IMPORT
#endif
```

The `Q_DECL_IMPORT` macro makes the compiler think that we're linking against a DLL, so all symbols are prefixed with `__imp_` and we get a bunch of "reference not found" errors. When building statically, QXmppGlobal.h must be changed to:

```cpp
#if defined(QXMPP_BUILD)
#    define QXMPP_EXPORT Q_DECL_EXPORT
#else
#    define QXMPP_EXPORT 
#endif
```

lnjX:
That's a problem indeed. For Kaidan we workarounded this by manually defining QXMPP_BUILD, but the right way would be to just check whether QXmpp was built static. Removing `Q_DECL_IMPORT` is not a good idea, I think.

leobasilio:
Sorry. I might have not expressed myself properly. Removing `Q_DECL_IMPORT` would break linking against the shared library. It's no good either. As far as I know, we can't tell on compilation time whether we're linking against a shared or a static library. When the linker kicks in, it's already too late. I was actually thinking about calling `sed` on install phase to rewrite that line, when building statically. But I just realized there's no `sed` on Windows... :thinking: Well, there must be some equivalent.

lnjX:
We can do that using CMake using configure_file (as we're already doing for other things like the version).

```cpp
#define QXMPP_STATIC @BUILD_STATIC@
#define QXMPP_SHARED @BUILD_SHARED@

#if QXMPP_STATIC
// ...
#endif
```

PR #299, opened 2020-05-17
Fixing QXmppGlobal.h for static linking

leobasilio:
Following the discussion on issue #297, here's my attempt to solve it.


lnjX:
Thanks for the PR. :)
I added a commit that makes this less hacky by adding a proper variable in cmake. Can you confirm that this still works all properly?

leobasilio:
Thanks for the fix. I'm really not that experienced with cmake. Just tried to be the least intrusive as possible. From what I can tell, it's working perfectly. :+1:

PR #303, opened 2020-07-11
 QXmppStanza::Error: Use std::optional<> internally 

lnjX:


CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ project(qxmpp VERSION 1.4.0)
 
 set(SO_VERSION 3)
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 17)
```
0xd34df00d, line 6: I'd probably make this a requirement:
```cmake
set(CMAKE_CXX_STANDARD_REQUIRED TRUE)
```

While we're on this, what about disabling the language extensions (like GNU ones on gcc/clang) to the C++ standard for the sake of portability?
```cmake
set(CMAKE_CXX_EXTENSIONS OFF)
```

src/base/QXmppStanza.h
```diff
@@ -187,14 +187,15 @@ class QXMPP_EXPORT QXmppStanza
...
-        /// \endcond
 
     private:
+        // ### QXmpp2: remove
         QString getConditionStr() const;
         void setConditionFromStr(const QString &cond);
 
         QString getTypeStr() const;
         void setTypeFromStr(const QString &type);
+        /// \endcond
```
0xd34df00d, line 198: This family of functions doesn't seem to affect neither API (it's private) nor ABI (it's not virtual after all), so looks like it's safe to remove it entirely without waiting for qxmpp 2.

lnjX: Can I really just remove a private method without affecting ABI compatibility?

lnjX: https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B says:
> You can...
> * remove private non-virtual functions if they are not called by any inline functions (and have never been).

Thank you, that's nice, didn't know that this was possible :)

CMakeLists.txt
lnjX: Yes, both points seem to be a good idea. :)

PR #312, opened 2020-09-29
If bind iq returns error, emit error signal

Bonnie080:
Hi, I found that some other xmpp servers can be configured to disallow the resource binding when there is resource conflict according to the 2nd behavior of https://tools.ietf.org/html/rfc6120#section-7.7.2.2. But QXmpp (as the client) does not handle that response at all. So I think it would be better to add just a few code to handle that situation.


lnjX:
Thanks for your pull-request. What does currently happen in that case? Could this cause any incompatibilities (probably not, but I'm just asking)? Apart from that this sounds reasonable and looks fine. :)

Bonnie080:
Currently if bind.type() !=QXmppIq::Result, I can't find any further handling. It seems that the client will just ignore that iq.

lnjX:
What will happen then, a timeout?

lnjX:
I actually can't reproduce this, `ejabberd` just closes my old connection and binds the new one (when I use the same resource). However I think I also had bind errors before (not with qxmpp, but with other xmpp libraries), I don't this change can break anything.

---

I had a look at the RFC, there are multiple ways a server can handle resource conflicts and there are also other errors.

The only problem I see here for QXmpp users is that, when you for example receive a "not-allowed" error, there's no way to find out that this is about the resource. And you can't even access the error string from the server. (However this is a general problem, not about this PR)

Bonnie080:
@lnjX 
Ah, sorry for the late reply. I was having a long holiday in China.
Yes the client won't do anything until the timeout.
ejabberd is one of the servers I've tried. It can be reproduced when the "resource_conflict" option is set to "closenew".
(But from the ejabberd doc, this option is deprecated and may be removed anytime in the future versions.)

PR #315, opened 2020-10-09
Enable QURL_NO_CAST_FROM_STRING+QT_NO_CAST_TO_ASCII and fix compilation

dfaure-kdab:
I have those always enabled, which is how I detected these issues.
This avoids QUrl/QString confusions, and QString/QByteArray confusions.


lnjX (approved):
Thanks for your patch!

PR #322, opened 2020-12-07
Fix UB in QXmppTransferManager when parsing stream features

lichtzeichner:
`featureForm()` returns a `QXmppDataForm` by value - directly taking a reference to a member of this class is undefined behavior and leads to a crash.

This PR changes the code to take references to both the form and the fields inside, thus extending the lifetime to the end of the function, fixing said crash.

### DIFF
--- a/src/client/QXmppTransferManager.cpp
+++ b/src/client/QXmppTransferManager.cpp
@@ -1342,7 +1342,9 @@ void QXmppTransferManager::streamInitiationResultReceived(const QXmppStreamIniti
         job->state() != QXmppTransferJob::OfferState)
         return;
 
-    const auto &fields = iq.featureForm().fields();
+    const auto &form = iq.featureForm();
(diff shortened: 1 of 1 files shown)


0xd34df00d:
Great catch, thanks!

The root cause seems to be that in `iq.featureForm().fields();`, the non-`const` overload of `fields()` is used which returns a (non-`const`) reference to the member, so the rules of lifetime extension don't apply.

0xd34df00d (approved):

lnjX (changes requested):
Great! Could you change the base branch of the PR to `1.3`? So if we create a v1.3.2 it'll be in there.

lichtzeichner:
@InjX: there you go. I hope I rebased the right way, but the diff looks good.

lnjX (approved):
Thank you very much! :)

PR #324, opened 2020-12-29
server: Add 'to' and 'from' attributes to dialback init stream header

Kaffeine:
The attributes are mandatory (see RFC 6120 paragraphs [8.1.1.2](https://tools.ietf.org/html/rfc6120#section-8.1.1.2) and [8.1.2.2](https://tools.ietf.org/html/rfc6120#section-8.1.2.2); also the attributes are listed in the example of stream header for [XEP-0220](https://xmpp.org/extensions/xep-0220.html#advertisement-traditional)).

This commit fixes S2S messaging with Prosody server (and probably many others).

### DIFF
--- a/src/server/QXmppOutgoingServer.cpp
+++ b/src/server/QXmppOutgoingServer.cpp
@@ -148,11 +148,14 @@ void QXmppOutgoingServer::handleStart()
     QXmppStream::handleStart();
 
     QString data = QString("<?xml version='1.0'?><stream:stream"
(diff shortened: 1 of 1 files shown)


Kaffeine:
Is there something _special_ I should do to convince codecov? :facepalm:

lnjX (approved):
Thanks for fixing this! :)

About codecov: I guess we can merge this little fix without a unit test.

PR #326, opened 2021-01-17
Avoid depending on gst/gst.h in a public header

0xd34df00d:
The gstreamer types used in QXmpp headers could just be forward-declared, so that the users of the library that don't care about gstreamer specifics can just not bother with gstreamer.

### DIFF
--- a/src/client/QXmppCallStream.h
+++ b/src/client/QXmppCallStream.h
@@ -26,11 +26,12 @@
 
 #include <QXmppGlobal.h>
 
-#include <gst/gst.h>
-
 #include <functional>
 
 #include <QObject>
+
+typedef struct _GstPad GstPad;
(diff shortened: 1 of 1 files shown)


JBBgameich:
[This](https://github.com/qxmpp-project/qxmpp/blob/master/src/CMakeLists.txt#L240) should probably be changed to `PRIVATE` so the exported CMake target doesn't depend on GStreamer anymore.

lnjX:
> [This](https://github.com/qxmpp-project/qxmpp/blob/master/src/CMakeLists.txt#L240) should probably be changed to `PRIVATE` so the exported CMake target doesn't depend on GStreamer anymore.

@0xd34df00d This sounds good ^

lnjX:
Replaced by #337.

PR #330, opened 2021-02-15
Fix strings in method documentation

melvo:


src/client/QXmppRosterManager.h
```diff
@@ -39,7 +39,7 @@ class QXmppRosterManagerPrivate;
 /// \brief The QXmppRosterManager class provides access to a connected client's
 /// roster.
 ///
-/// \note It's object should not be created using it's constructor. Instead
+/// \note Its object should not be created using its constructor. Instead
```
lnjX, line 42: This is also wrong in some other extensions, could you also fix it there?

lnjX (approved):

PR #331, opened 2021-02-16
QXmppRosterManager: Request entries with MIX annotation

melvo:


src/client/QXmppRosterManager.cpp
```diff
@@ -121,6 +121,8 @@ void QXmppRosterManager::_q_connected()
         QXmppRosterIq roster;
         roster.setType(QXmppIq::Get);
         roster.setFrom(client()->configuration().jid());
+        roster.setMixAnnotate(true);
```
lnjX, line 124:
```suggestion
        // TODO: Check whether the server actually supports MIX
        roster.setMixAnnotate(true);
```

lnjX (approved):

PR #332, opened 2021-02-17
QXmppRosterManager: Add method for retrieving MIX channel entry

melvo:


lnjX (commented):
Are you sure that the participant ID should be globally unique? I think the XEP only says something about the uniqueness per channel, so theoretically the user could have got the same participant id in different channels.

I mean clients may still search for the roster item by participant id, but we don't need to support that by adding an extra method for it, do we?

melvo:
> Are you sure that the participant ID should be globally unique? I think the XEP only says something about the uniqueness per channel, so theoretically the user could have got the same participant id in different channels.
You are right. I introduced it because I could not find a way to determine the channel JID. But maybe that method is not needed anymore.

> I mean clients may still search for the roster item by participant id, but we don't need to support that by adding an extra method for it, do we?
This pull request is closed because I will check if searching for a channel JID by a participant ID is really needed.

PR #339, opened 2021-03-17
Fix build with Qt 5.7

lnjX:
We only test 5.9 with the CI... :/

### DIFF
--- a/src/base/QXmppGlobal.h.in
+++ b/src/base/QXmppGlobal.h.in
@@ -73,5 +73,9 @@ inline QLatin1String QXmppVersion()
 #define Q_DECL_ENUMERATOR_DEPRECATED_X(msg)
 #endif
 
+#ifndef QT_WARNING_DISABLE_DEPRECATED
+#define QT_WARNING_DISABLE_DEPRECATED
+#endif
+
 #endif // QXMPPGLOBAL_H
 
--- a/src/base/QXmppUtils.cpp
+++ b/src/base/QXmppUtils.cpp
@@ -126,8 +126,13 @@ QDateTime QXmppUtils::datetimeFromString(const QString &str)
 ///
 QString QXmppUtils::datetimeToString(const QDateTime &dt)
 {
-    if (dt.time().msec())
+    if (dt.time().msec()) {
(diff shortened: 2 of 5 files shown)


tehnick:
Why do you care about Qt 5.7.x?

lnjX:
Officially we still support 5.7 (and I came up with the idea of testing that). We could also just increase our dependency to 5.9.

tehnick:
> Officially we still support 5.7

I see only one possible reason for this: support of MS Windows XP. Which is unsupported by Microsoft since 2014...

> We could also just increase our dependency to 5.9.

+1 for this. Since the next minor version.

Just a reminder... Classical versioning scheme: `<major>.<minor>.<patch>`.

tehnick:
Small reminder 2: we still support Ubuntu 18.04 LTS, where Qt 5.9.x is used:
https://packages.ubuntu.com/bionic/qtbase5-dev
So the minimal requirement to Qt version as 5.9 will not break this.

lnjX:
> I see only one possible reason for this: support of MS Windows XP. Which is unsupported by Microsoft since 2014...
I think that even required 5.6. But I don't think it works with C++17 anyways.

> Just a reminder... Classical versioning scheme: `<major>.<minor>.<patch>`.
What do you mean? (I suggested 5.9 because of Ubuntu Bionic.)

tehnick:
>> We could also just increase our dependency to 5.9.
>
> +1 for this. Since the next minor version.

Please add an issue to not forget this.

lnjX:
I can just increase it on the master branch already

tehnick:
>>  I see only one possible reason for this: support of MS Windows XP. Which is unsupported by Microsoft since 2014...
>
> I think that even required 5.6. But I don't think it works with C++17 anyways.

I still may build `qxmpp` for WinXP using one of snapshots of [MXE project](https://mxe.cc/) with Qt 5.7.1.

> What do you mean?

I mean:
1) save support of Qt 5.7.x in qxmpp 1.4.x
2) bump minimal requirement to Qt 5.9.x since qxmpp 1.5.0 release

tehnick:
WinXP support was broken in Qt only since Qt 5.8.0 if you do not remember.

lnjX:
Alright, I created #340.

PR #346, opened 2021-04-23
RosterManager: Add signal emitting status of subscription request

melvo:


src/client/QXmppRosterManager.cpp
```diff
@@ -234,6 +234,7 @@ void QXmppRosterManager::_q_presenceReceived(const QXmppPresence &presence)
             subscribe(bareJid);
         } else {
             emit subscriptionReceived(bareJid);
+            emit subscriptionRequestReceived(bareJid, presence.statusText());
```
lnjX, line 237: To make this future proof, let's include the QXmppPresence and a unit test would be nice of course. :)

src/client/QXmppRosterManager.h
```diff
@@ -119,6 +119,17 @@ public Q_SLOTS:
...
 
+    /// This signal is emitted when a contact asks to subscribe to your presence.
+    ///
+    /// You can either accept the request by calling acceptSubscription() or refuse it
+    /// by calling refuseSubscription().
+    ///
+    /// \note If you set QXmppConfiguration::autoAcceptSubscriptions() to true, this
+    /// signal will not be emitted.
+    ///
+    /// \since QXmpp 1.5
```
lnjX, line 130: Can you move this into the cpp file (like here: https://github.com/qxmpp-project/qxmpp/blob/master/src/client/QXmppAttentionManager.cpp#L58-L77)

lnjX (changes requested):

src/client/QXmppRosterManager.cpp
```diff
@@ -31,6 +32,23 @@
 
 #include <QDomElement>
 
+///
+/// \fn QXmppRosterManager::subscriptionRequestReceived
+///
+/// This signal is emitted when a JID asks to subscribe to the user's presence.
+///
+/// The suer can either accept the request by calling acceptSubscription() or refuse it
```
lnjX, line 40:
```suggestion
/// The user can either accept the request by calling acceptSubscription() or refuse it
```

PR #347, opened 2021-04-28
Add QFuture support

### LINKED ISSUE #96: Handle IQ responses properly with QFutures

lnjX:
Closes #96.

### DIFF
--- a/src/client/QXmppEntityTimeManager.h
+++ b/src/client/QXmppEntityTimeManager.h
@@ -26,28 +26,37 @@
 
 #include "QXmppClientExtension.h"
 
+#include <variant>
+
+template<class T>
+class QFuture;
 class QXmppEntityTimeIq;
 
+///
 /// \brief The QXmppEntityTimeManager class provided the functionality to get
 /// the local time of an entity as defined by \xep{0202}: Entity Time.
 ///
 /// \ingroup Managers
-
+///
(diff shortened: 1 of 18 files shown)


src/client/QXmppEntityTimeManager.h
```diff
@@ -26,28 +26,37 @@
...
+///
 class QXMPP_EXPORT QXmppEntityTimeManager : public QXmppClientExtension
 {
     Q_OBJECT
 
 public:
-    QString requestTime(const QString& jid);
+    QString requestTime(const QString &jid);
+
+    using EntityTimeResult = std::variant<QXmppEntityTimeIq, QXmpp::PacketState>;
```
lnjX, line 48: should probably include `QXmppStanza::Error` for consistency and so the library user doesn't need to do the check.

lnjX:
@0xd34df00d It would be great if you could have a quick look at this as it adds a new API concept. My open questions are (a) is the new `QXmpp` namespace a good idea? and (b) how can we properly test the new APIs with QFutures as they require an real connection.

PR #348, opened 2021-04-29
MUC History Management

wx984490083:
add history element into presence for MUC
ref https://xmpp.org/extensions/xep-0045.html#enter-history


src/client/QXmppMucManager.h
```diff
@@ -183,6 +187,10 @@ class QXMPP_EXPORT QXmppMucRoom : public QObject
     QString subject() const;
     void setSubject(const QString &subject);
 
+    // ref https://xmpp.org/extensions/xep-0045.html#enter-managehistory
+    QDateTime historySince() const;
+    void setHistorySince(const QDateTime& since);
+
```
lnjX, line 193: I see two problems here.
* There are three other options available in the XEP for limiting the output. Adding a getter/setter here *and* in QXmppPresence is not a good idea. Can you add a class for all of the options? (e.g. `QXmppMucHistoryLimit`)
* The attributes should not be saved permanently in the room object. A parameter for it in the join function is enough.

lnjX (changes requested):
Thank you for your pull-request so far! Some changes are still needed and also unit testing for at least the data classes (QXmppPresence) would be important. :)

lnjX:
I'll close this due to inactivity. If there's still interest in this or you need help, just say so.

PR #350, opened 2021-05-12
/// Sends a message to the room. Use the parameter msg to send it.

NathanJamesBlackerby:
///
/// msg's To, Type, and Body will be set by this function.
///
/// \return true if the request was sent, false otherwise

bool QXmppMucRoom::sendMessage(const QString &text, QXmppMessage &msg)
{
    msg.setTo(d->jid);
    msg.setType(QXmppMessage::GroupChat);
    msg.setBody(text);
    return d->client->sendPacket(msg);
}

### DIFF
--- a/src/client/QXmppMucManager.cpp
+++ b/src/client/QXmppMucManager.cpp
@@ -343,6 +343,20 @@ bool QXmppMucRoom::sendMessage(const QString &text)
     return d->client->sendPacket(msg);
 }
 
+/// Sends a message to the room. Use the parameter msg to send it.
+///
+/// msg's To, Type, and Body will be set by this function.
+///
+/// \return true if the request was sent, false otherwise
+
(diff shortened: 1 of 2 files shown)


NathanJamesBlackerby:
I needed control over the message Id. This was not possible without making my own MUC manager and MUCroom or by adding this small function to the project. The original sendMessage function would create a new QXmppMessage with default member values and use that to send the message. I needed to set the message id before sending the text for my project.

lnjX:
If you need control over the message you can just create your own message and directly send it via QXmppClient::sendPacket.

lnjX:
https://github.com/qxmpp-project/qxmpp/commit/953d8588b94fec779bdf667c024150c9690470c4
I improved the documentation on this.

NathanJamesBlackerby:
I was able to update my code with what you told me thank you


Nathan J. Blackerby
AMEWAS


________________________________

PR #384, opened 2021-12-08
Update to XEP-0380 version 0.4

melvo:
- [ ] Documentation:

### DIFF
--- a/src/base/QXmppMessage.cpp
+++ b/src/base/QXmppMessage.cpp
@@ -70,7 +70,9 @@ static const QStringList ENCRYPTION_NAMESPACES = {
     ns_otr,
     ns_legacy_openpgp,
     ns_ox,
-    ns_omemo
+    ns_omemo,
+    ns_omemo_1,
+    ns_omemo_2
 };
 
 static const QStringList HINT_TYPES = {
@@ -86,7 +88,9 @@ static const QStringList ENCRYPTION_NAMES = {
     QStringLiteral("OTR"),
     QStringLiteral("Legacy OpenPGP"),
     QStringLiteral("OpenPGP for XMPP (OX)"),
-    QStringLiteral("OMEMO")
+    QStringLiteral("OMEMO"),
+    QStringLiteral("OMEMO 1"),
+    QStringLiteral("OMEMO 2")
 };
 
 static bool checkElement(const QDomElement &element, const QString &tagName, const QString &xmlns)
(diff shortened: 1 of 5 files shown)


src/base/QXmppMessage.h
```diff
@@ -99,17 +99,19 @@ class QXMPP_EXPORT QXmppMessage : public QXmppStanza
...
-        OTR,                ///< \xep{0364}: Current Off-the-Record Messaging Usage
-        LegacyOpenPGP,      ///< \xep{0027}: Current Jabber OpenPGP Usage
-        OX,                 ///< \xep{0373}: OpenPGP for XMPP
-        OMEMO               ///< \xep{0384}: OMEMO Encryption
+        OTR,                ///< \xep{0364, Current Off-the-Record Messaging Usage}
+        LegacyOpenPGP,      ///< \xep{0027, Current Jabber OpenPGP Usage}
+        OX,                 ///< \xep{0373, OpenPGP for XMPP}
+        OMEMO,              ///< \xep{0384, OMEMO Encryption}
+        OMEMO1,             ///< \xep{0384, OMEMO Encryption} since version 0.4
+        OMEMO2              ///< \xep{0384, OMEMO Encryption} since version 0.8
```
lnjX, line 114: needs to be added in .cpp too

src/base/QXmppMessage.cpp
```diff
@@ -86,7 +88,9 @@ static const QStringList ENCRYPTION_NAMES = {
     QStringLiteral("OTR"),
     QStringLiteral("Legacy OpenPGP"),
     QStringLiteral("OpenPGP for XMPP (OX)"),
-    QStringLiteral("OMEMO")
+    QStringLiteral("OMEMO"),
+    QStringLiteral("OMEMO since version 0.4"),
+    QStringLiteral("OMEMO since version 0.8")
```
lnjX, line 93:
```suggestion
    QStringLiteral("OMEMO:1"),
    QStringLiteral("OMEMO:2")
```
This is the name as it should be displayed in the UI, so a short name would be good. "OMEMO v0.8+" would also be an option, but I think the XEP versioning scheme isn't so good here, because the 0.x version number also changes when the encryption stays compatible.

lnjX: Okay, no actually this must be the name from the XEP, so "OMEMO 1" and "OMEMO 2".

lnjX (approved):

PR #386, opened 2022-01-08
Run 'clang-format' on all files

melvo:
- [ ] Documentation:

### DIFF
--- a/src/client/QXmppClient.cpp
+++ b/src/client/QXmppClient.cpp
@@ -30,7 +30,7 @@
 using namespace QXmpp::Private;
 
 /// \cond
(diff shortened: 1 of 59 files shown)


src/client/QXmppClient.cpp
```diff
@@ -266,7 +266,7 @@ bool QXmppClient::removeExtension(QXmppClientExtension* extension)
 ///
 /// \since QXmpp 1.5
 ///
-QXmppE2eeExtension *QXmppClient::encryptionExtension() const
+QXmppE2eeExtension* QXmppClient::encryptionExtension() const
```
lnjX, line 269: this is not what we want... don't know why clang-format does this

PR #388, opened 2022-01-10
Refactor trust storages

melvo:
QXmppTrustStorage is now the base class for all trust storages used by
end-to-end encryption managers.
QXmppAtmTrustStorage is used by QXmppAtmManager.

QXmppTrustMemoryStorage is now the base class for all trust storages
that use the memory for storing data.
QXmppAtmTrustMemoryStorage can be used by QXmppAtmManager.

Methods needed by the upcoming OMEMO implementation are added.
Some existing methods are refactored.

- [x] Documentation:

### DIFF
--- a/src/client/QXmppAtmTrustMemoryStorage.cpp
+++ b/src/client/QXmppAtmTrustMemoryStorage.cpp
@@ -1,0 +1,145 @@
+// SPDX-FileCopyrightText: 2022 Melvin Keskin <melvo@olomono.de>
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+#include "QXmppAtmTrustMemoryStorage.h"
+
+#include "QXmppFutureUtils_p.h"
+#include "QXmppTrustMessageKeyOwner.h"
+
+using namespace QXmpp::Private;
+
+///
+/// \class QXmppAtmTrustMemoryStorage
+///
+/// \brief The QXmppAtmTrustMemoryStorage class stores trust data for
+/// \xep{0450, Automatic Trust Management (ATM)} in the memory.
+///
+/// \warning THIS API IS NOT FINALIZED YET!
+///
+/// \since QXmpp 1.5
+///
+
+struct UnprocessedKey
+{
+    QByteArray id;
+    QString ownerJid;
+    QByteArray senderKeyId;
+    bool trust;
+};
+
+class QXmppAtmTrustMemoryStoragePrivate
+{
+public:
+    // encryption protocols mapped to trust message data received from endpoints
+    // with unauthenticated keys
+    QMultiHash<QString, UnprocessedKey> keys;
+};
+
+///
+/// Constructs an ATM trust memory storage.
+///
+QXmppAtmTrustMemoryStorage::QXmppAtmTrustMemoryStorage()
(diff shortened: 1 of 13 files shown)


tests/qxmppatmtrustmemorystorage/tst_qxmppatmtrustmemorystorage.cpp
```diff
@@ -0,0 +1,317 @@
...
+ */
+
+#include "QXmppAtmTrustMemoryStorage.h"
+#include "QXmppConstants.cpp"
+#include "QXmppConstants_p.h"
+#include "QXmppTrustMessageKeyOwner.h"
+
+#include "util.h"
+
+class tst_QXmppAtmTrustMemoryStorage : public QObject
```
lnjX, line 31: It would be nice to have this in the same test class as trustmemorystorage, because the unit tests really slow down building.

src/client/QXmppAtmTrustMemoryStorage.cpp
```diff
@@ -0,0 +1,164 @@
...
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ */
```
lnjX, line 22: now needs to be replaced with reuse header

src/client/QXmppTrustStorage.h
```diff
@@ -20,43 +18,44 @@ class QXMPP_EXPORT QXmppTrustStorage
...
     virtual ~QXmppTrustStorage() = default;
 
-    virtual QFuture<void> setSecurityPolicies(const QString &encryption = {}, SecurityPolicy securityPolicy = SecurityPolicy::NoSecurityPolicy) = 0;
+    virtual QFuture<void> setSecurityPolicy(const QString &encryption, SecurityPolicy securityPolicy = SecurityPolicy::NoSecurityPolicy) = 0;
     virtual QFuture<SecurityPolicy> securityPolicy(const QString &encryption) = 0;
 
-    virtual QFuture<void> addOwnKey(const QString &encryption, const QByteArray &keyId) = 0;
-    virtual QFuture<void> removeOwnKey(const QString &encryption) = 0;
+    virtual QFuture<void> setOwnKey(const QString &encryption, const QByteArray &keyId = {}) = 0;
     virtual QFuture<QByteArray> ownKey(const QString &encryption) = 0;
```
lnjX, line 44: Currently `setOwnKey("encryption:xml-ns");` would reset the key, which looks a bit weird. How about this?:
```suggestion
    virtual QFuture<QByteArray> ownKey(const QString &encryption) = 0;
    virtual QFuture<void> setOwnKey(const QString &encryption, const QByteArray &keyId) = 0;
    inline QFuture<void> resetOwnKey(const QString &encryption) { return setOwnKey(encryption, QByteArray()); }
```

PR #390, opened 2022-01-23
Add QXmppE2eeMetadata used by QXmppStanza

melvo:
That class contains addtional data needed for end-to-end encryption purposes.
'senderKey' is moved from QXmppMessage to QXmppE2eeMetadata.
'sceTimestamp' is introduced.

- [x] Documentation:

### DIFF
--- a/src/base/QXmppStanza.h
+++ b/src/base/QXmppStanza.h
@@ -8,6 +8,8 @@
 #ifndef QXMPPSTANZA_H
 #define QXMPPSTANZA_H
 
+#include <optional>
+
 #include <QByteArray>
 #include <QSharedData>
 #include <QString>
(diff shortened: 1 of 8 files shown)


src/base/QXmppStanza.h
```diff
@@ -63,6 +66,28 @@ class QXMPP_EXPORT QXmppExtendedAddress
...
+    ~QXmppE2eeMetadata();
+
+    QXmppE2eeMetadata &operator=(const QXmppE2eeMetadata &other);
+
+    QByteArray senderKey() const;
+    void setSenderKey(const QByteArray &keyId);
+
+    // XEP-0420: Stanza Content Encryption
+    std::optional<QDateTime> sceTimestamp() const;
+    void setSceTimestamp(const std::optional<QDateTime> &timestamp);
```
lnjX, line 85: QDateTime itself is already nullable. I think that's already enough. I wouldn't use `std::optional<QString>` either.

PR #392, opened 2022-01-27
PubSub / PEP improvements

melvo:
- [ ] Documentation:

### DIFF
--- a/src/base/QXmppConstants.cpp
+++ b/src/base/QXmppConstants.cpp
@@ -41,9 +41,18 @@ const char *ns_vcard = "vcard-temp";
 const char *ns_rsm = "http://jabber.org/protocol/rsm";
 // XEP-0060: Publish-Subscribe
 const char *ns_pubsub = "http://jabber.org/protocol/pubsub";
+const char *ns_pubsub_auto_create = "http://jabber.org/protocol/pubsub#auto-create";
+const char *ns_pubsub_config_node = "http://jabber.org/protocol/pubsub#config-node";
+const char *ns_pubsub_config_node_max = "http://jabber.org/protocol/pubsub#config-node-max";
+const char *ns_pubsub_create_and_configure = "http://jabber.org/protocol/pubsub#create-and-configure";
+const char *ns_pubsub_create_nodes = "http://jabber.org/protocol/pubsub#create-nodes";
 const char *ns_pubsub_errors = "http://jabber.org/protocol/pubsub#errors";
 const char *ns_pubsub_event = "http://jabber.org/protocol/pubsub#event";
+const char *ns_pubsub_multi_items = "http://jabber.org/protocol/pubsub#multi-items";
+const char *ns_pubsub_node_config = "http://jabber.org/protocol/pubsub#node_config";
 const char *ns_pubsub_owner = "http://jabber.org/protocol/pubsub#owner";
+const char *ns_pubsub_publish = "http://jabber.org/protocol/pubsub#publish";
+const char *ns_pubsub_publish_options = "http://jabber.org/protocol/pubsub#publish-options";
 const char *ns_pubsub_rsm = "http://jabber.org/protocol/pubsub#rsm";
 // XEP-0065: SOCKS5 Bytestreams
 const char *ns_bytestreams = "http://jabber.org/protocol/bytestreams";
--- a/src/client/QXmppPubSubManager.cpp
+++ b/src/client/QXmppPubSubManager.cpp
@@ -54,7 +54,9 @@ using namespace QXmpp::Private;
 ///    Requesting a continuation
 ///  - Requesting most recent items (max_items=x):
 ///    https://xmpp.org/extensions/xep-0060.html#subscriber-retrieve-requestrecent
-///  - (Manually) subscribing to a node
+///  - subscribe()/unsubscribe():
+///    - return subscription on success
+///    - correctly handle configuration required (and other) cases
 ///
 /// \ingroup Managers
 ///
--- a/src/client/QXmppPubSubManager.cpp
+++ b/src/client/QXmppPubSubManager.cpp
@@ -741,6 +741,56 @@ QFuture<QXmppPubSubManager::Result> QXmppPubSubManager::unsubscribeFromNode(cons
 ///
 
 ///
+/// \fn QXmppPubSubManager::requestPepItem(const QString &nodeName, const QString &itemId)
+///
+/// Requests a specific item of a PEP node.
+///
+/// This is a convenience method equivalent to calling
+/// QXmppPubSubManager::requestItem on the current account's bare JID.
+///
+/// \param nodeName name of the PEP node whose item is requested
+/// \param itemId ID of the requested item
+///
+
+///
+/// \fn QXmppPubSubManager::requestPepItem(const QString &nodeName, StandardItemId itemId)
+///
+/// Requests a specific item of a PEP node.
+///
+/// The default value of itemId is used for singleton nodes (i.e., the node's
+/// single item is requested).
+///
+/// This is a convenience method equivalent to calling
+/// QXmppPubSubManager::requestItem on the current account's bare JID.
+///
+/// \param nodeName name of the PEP node whose item is requested
+/// \param itemId ID of the requested item
+///
+
+///
+/// \fn QXmppPubSubManager::requestPepItems(const QString &nodeName)
+///
+/// Requests all items of a PEP node.
+///
+/// This is a convenience method equivalent to calling
+/// QXmppPubSubManager::requestItems on the current account's bare JID.
+///
+/// \param nodeName name of the PEP node whose items are requested
+///
+
+///
+/// \fn QXmppPubSubManager::requestPepItemIds(const QString &nodeName)
+///
+/// Requests the IDs of all items of a pubsub service node via service
+/// discovery.
+///
+/// This is a convenience method equivalent to calling
+/// QXmppPubSubManager::requestItemIds on the current account's bare JID.
+///
+/// \param nodeName name of the PEP node whose item IDs are requested
+///
+
+///
 /// \fn QXmppPubSubManager::retractPepItem(const QString &nodeName, const QString &itemId)
 ///
 /// Deletes an item from a PEP node.
--- a/src/client/QXmppPubSubManager.cpp
+++ b/src/client/QXmppPubSubManager.cpp
@@ -643,6 +643,46 @@ QFuture<QXmppPubSubManager::Result> QXmppPubSubManager::cancelNodeConfiguration(
 }
 
 ///
+/// Subscribes to a node.
+///
+/// \warning THIS API IS NOT FINALIZED YET!
+///
+/// \param serviceJid JID of the pubsub service
+/// \param nodeName name of the pubsub node being subscribed
+/// \param subscriberJid bare or full JID of the subscriber
+///
+QFuture<QXmppPubSubManager::Result> QXmppPubSubManager::subscribeToNode(const QString &serviceJid, const QString &nodeName, const QString &subscriberJid)
+{
+    QXmppPubSubIq request;
+    request.setType(QXmppIq::Set);
+    request.setTo(serviceJid);
+    request.setQueryNode(nodeName);
+    request.setQueryType(QXmppPubSubIq<>::Subscribe);
+    request.setQueryJid(subscriberJid);
+    return client()->sendGenericIq(std::move(request));
+}
+
+///
+/// Unsubscribes from a node.
+///
+/// \warning THIS API IS NOT FINALIZED YET!
+///
+/// \param serviceJid JID of the pubsub service
+/// \param nodeName name of the pubsub node being subscribed
+/// \param subscriberJid bare or full JID of the subscriber
+///
+QFuture<QXmppPubSubManager::Result> QXmppPubSubManager::unsubscribeFromNode(const QString &serviceJid, const QString &nodeName, const QString &subscriberJid)
+{
+    QXmppPubSubIq request;
+    request.setType(QXmppIq::Set);
+    request.setTo(serviceJid);
+    request.setQueryNode(nodeName);
+    request.setQueryType(QXmppPubSubIq<>::Unsubscribe);
+    request.setQueryJid(subscriberJid);
+    return client()->sendGenericIq(std::move(request));
+}
+
+///
 /// \fn QFuture<Result> QXmppPubSubManager::createPepNode(const QString &nodeName)
 ///
 /// Creates an empty PEP node with the default configuration.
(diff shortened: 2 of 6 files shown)


src/client/QXmppPubSubManager.cpp
```diff
@@ -305,6 +344,35 @@ auto QXmppPubSubManager::deleteNode(const QString &jid, const QString &nodeName)
...
+
+    return chainIq(client()->sendIq(std::move(request)), this, [](QXmppDiscoveryIq &&iq) -> ItemIdsResult {
+        const auto queryItems = iq.items();
+        QVector<QString> itemIds;
+        for (const auto &queryItem : queryItems) {
+            itemIds << queryItem.name();
+        }
+        return itemIds;
+    });
+}
```
lnjX, line 374: This requests all items with their contents and so hides the overhead from the user. This can lead to misuse (i.e. user requests list of items and then requests some of the items separately although the items were already received).

I don't see a necessity for this function as the user can just request all items. Iterating over or extracting the IDs from the result is trivial then. Then the user also knows what they do and which overhead they create.

Also, the item IDs can already be requested via the discovery manager. This is probably the preferred option anyways. (see https://xmpp.org/extensions/xep-0060.html#entity-discoveritems)

melvo: The content is not requested because it uses what you suggested (https://xmpp.org/extensions/xep-0060.html#entity-discoveritems). Have a look at `tst_QXmppPubSubManager::testRequestItemIds()`.

Should `requestItemIds()` use `QXmppDiscoveryManager::requestDiscoItems()`?

lnjX: Sorry, forget what I said, nvm.

src/client/QXmppPubSubManager.h
```diff
@@ -20,13 +20,19 @@
...
-extern const char *PUBSUB_CURRENT_ITEM;
-
 class QXMPP_EXPORT QXmppPubSubManager : public QXmppClientExtension
 {
     Q_OBJECT
 
 public:
+    enum ItemId {
+        Current  ///< Item of a singleton node
+    };
```
lnjX, line 39: the enum itself needs to be documented

src/client/QXmppPubSubManager.h
```diff
@@ -20,13 +20,19 @@
...
 class QXmppPubSubSubscribeOptions;
 
-extern const char *PUBSUB_CURRENT_ITEM;
-
 class QXMPP_EXPORT QXmppPubSubManager : public QXmppClientExtension
 {
     Q_OBJECT
 
 public:
+    enum ItemId {
```
lnjX, line 37:
```suggestion
    enum StandardItemId {
```

src/client/QXmppPubSubManager.h
```diff
@@ -20,13 +20,19 @@
...
+        Current  ///< Item of a singleton node
+    };
+
+    const QVector<QString> ITEM_IDS = {
+        QStringLiteral("current")
+    };
+
```
lnjX, line 44: this shouldn't be public and if you add it here it needs to be static (otherwise it would affect the size of QXmppPubSubManager)

I'd suggest to do:
```
private:
    static QString standardItemIdToString(StandardItemId);
```
(and then defining it in the .cpp file)

src/client/QXmppPubSubManager.h
```diff
@@ -52,14 +60,18 @@ class QXMPP_EXPORT QXmppPubSubManager : public QXmppClientExtension
...
     QFuture<Result> createNode(const QString &jid, const QString &nodeName, const QXmppPubSubNodeConfig &config);
     QFuture<InstantNodeResult> createInstantNode(const QString &jid);
     QFuture<InstantNodeResult> createInstantNode(const QString &jid, const QXmppPubSubNodeConfig &config);
     QFuture<Result> deleteNode(const QString &jid, const QString &nodeName);
+    QFuture<ItemIdsResult> requestItemIds(const QString &serviceJid, const QString &nodeName);
+    template<typename T = QXmppPubSubItem>
+    QFuture<ItemResult<T>> requestItem(const QString &jid, const QString &nodeName, const QString &itemId);
     template<typename T = QXmppPubSubItem>
-    QFuture<ItemResult<T>> requestItem(const QString &jid, const QString &nodeName, const QString &itemId = PUBSUB_CURRENT_ITEM);
+    QFuture<ItemResult<T>> requestItem(const QString &jid, const QString &nodeName, ItemId itemId = Current);
```
lnjX, line 83: i'm not 100%-sure about the default parameter here, it's probably very convenient, but I could also imagine cases where it's not so clear what a `requestItem(jid, node)` call does.

src/base/QXmppConstants.cpp
```diff
@@ -1,4 +1,5 @@
 // SPDX-FileCopyrightText: 2009 Manjeet Dahiya <manjeetdahiya@gmail.com>
+// SPDX-FileCopyrightText: 2022 Melvin Keskin <melvo@olomono.de>
```
lnjX, line 2: not needed in this file / for this type of changes

src/client/QXmppPubSubManager.cpp
```diff
@@ -20,6 +21,8 @@
 
 using namespace QXmpp::Private;
 
+const char *PUBSUB_CURRENT_ITEM = "current";
+
```
lnjX, line 25:
```suggestion
```

src/client/QXmppPubSubManager.cpp
```diff
@@ -160,6 +163,40 @@ QXmppPubSubManager::~QXmppPubSubManager()
...
+///
+/// This uses a \xep{0030, Service Discovery} info request to get a list of
+/// features.
+///
+/// \param serviceJid JID of the entity hosting the pubsub service
+/// \param onlyForPepService whether to retrieve features only if the pubsub
+///        service is a PEP service
+/// \return
+///
+QFuture<QXmppPubSubManager::FeaturesResult> QXmppPubSubManager::requestFeatures(const QString &serviceJid, const bool onlyForPepService)
```
lnjX: The documentation is missing:
 * the function checks that the service has the correct identity
 * what happens / is returned when the service doesn't have the correct identity

src/client/QXmppPubSubManager.cpp
```diff
@@ -160,6 +163,50 @@ QXmppPubSubManager::~QXmppPubSubManager()
...
+                    return identityType == QStringLiteral("service");
+                case Pep:
+                    return identityType == QStringLiteral("pep");
+                }
+            }
+            return false;
+        });
+
+        if (isPubSubServiceFound) {
+            return iq.features().toVector();
```
lnjX, line 203:
```suggestion
#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
            return iq.features();
#else
            return iq.features().toVector();
#endif
```
(toVector() has been removed in Qt 6)

src/client/QXmppPubSubManager.cpp
```diff
@@ -302,6 +349,35 @@ auto QXmppPubSubManager::deleteNode(const QString &jid, const QString &nodeName)
...
+{
+    QXmppDiscoveryIq request;
+    request.setType(QXmppIq::Get);
+    request.setQueryType(QXmppDiscoveryIq::ItemsQuery);
+    request.setQueryNode(nodeName);
+    request.setTo(serviceJid);
+
+    return chainIq(client()->sendIq(std::move(request)), this, [](QXmppDiscoveryIq &&iq) -> ItemIdsResult {
+        const auto queryItems = iq.items();
+        QVector<QString> itemIds;
```
lnjX, line 373:
```suggestion
        QVector<QString> itemIds;
        itemIds.reserve(queryItems.size());
```

lnjX (changes requested):

src/client/QXmppPubSubManager.h
melvo: I think that it is a good default because there are several use cases where one node does only contain one item.

src/client/QXmppPubSubManager.cpp
melvo: Removed by 4711fe94c376e60744fb05c9e71563a958072db2

src/client/QXmppPubSubManager.h
lnjX: OK

src/client/QXmppPubSubManager.cpp
```diff
@@ -21,7 +21,9 @@
 
 using namespace QXmpp::Private;
 
-const char *PUBSUB_CURRENT_ITEM = "current";
+const QVector<QString> ITEM_IDS = {
```
lnjX, line 24:
```suggestion
static const QVector<QString> ITEM_IDS = {
```

src/client/QXmppPubSubManager.cpp
```diff
@@ -160,6 +165,57 @@ QXmppPubSubManager::~QXmppPubSubManager()
...
+/// This uses a \xep{0030, Service Discovery} info request to get a list of
+/// features.
+///
+/// Only if the service is of type serviceType, its features are returned.
+/// Otherwise, no features are returned.
+///
+/// \param serviceJid JID of the entity hosting the pubsub service
+/// \param serviceType type of service to retrieve features for
+/// \return
+///
```
lnjX, line 180: My suggestion: :)
```suggestion
///
/// Checks the identities and returns all features of a pubsub service and via
/// service discovery.
///
/// This uses a \xep{0030, Service Discovery} info request to get the service
/// identities and a list of features.
///
/// If the service is <i>not</i> of type \c serviceType, <i>no</i> features are
/// returned.
///
/// \param serviceJid JID of the entity hosting the pubsub service
/// \param serviceType service type that the identities are checked for
///
```

src/client/QXmppPubSubManager.cpp
```diff
@@ -589,6 +691,68 @@ QFuture<QXmppPubSubManager::Result> QXmppPubSubManager::cancelNodeConfiguration(
...
+    request.setType(QXmppIq::Set);
+    request.setTo(serviceJid);
+    request.setQueryNode(nodeName);
+    request.setQueryType(QXmppPubSubIq<>::Unsubscribe);
+    request.setQueryJid(subscriberJid);
+    return client()->sendGenericIq(std::move(request));
+}
+
+///
+/// \fn QFuture<Result> QXmppPubSubManager::requestPepFeatures()
```
lnjX, line 735: return type seems to be wrong

src/client/QXmppPubSubManager.cpp
```diff
@@ -589,6 +691,68 @@ QFuture<QXmppPubSubManager::Result> QXmppPubSubManager::cancelNodeConfiguration(
...
+/// Requests all features of the own PEP service via service discovery.
+///
+/// This is a convenience method equivalent to calling
+/// QXmppPubSubManager::requestFeatures on the current account's bare JID.
+///
+/// \return
+///
+
+///
+/// \fn QFuture<Result> QXmppPubSubManager::fetchPepNodes()
```
lnjX, line 746:
```suggestion
/// \fn QFuture<NodesResult> QXmppPubSubManager::fetchPepNodes()
```

src/client/QXmppPubSubManager.cpp
```diff
@@ -635,11 +799,55 @@ QFuture<QXmppPubSubManager::Result> QXmppPubSubManager::cancelNodeConfiguration(
 /// \return
 ///
 
+///
+/// \fn QXmppPubSubManager::requestPepItem(const QString &nodeName, const QString &itemId = PUBSUB_CURRENT_ITEM)
```
lnjX, line 803: needs to be updated

src/client/QXmppPubSubManager.cpp
```diff
@@ -636,10 +813,80 @@ QFuture<QXmppPubSubManager::Result> QXmppPubSubManager::cancelNodeConfiguration(
 ///
 
 ///
-/// \fn QXmppPubSubManager::retractPepItem
+/// \fn requestPepItem(const QString &nodeName, const QString &itemId)
```
melvo, line 816:
```suggestion
/// \fn QXmppPubSubManager::requestPepItem(const QString &nodeName, const QString &itemId)
```

src/client/QXmppPubSubManager.cpp
```diff
@@ -636,10 +813,80 @@ QFuture<QXmppPubSubManager::Result> QXmppPubSubManager::cancelNodeConfiguration(
...
+/// This is a convenience method equivalent to calling
+/// QXmppPubSubManager::requestItem on the current account's bare JID.
+///
+/// \param nodeName name of the PEP node whose item is requested
+/// \param itemId ID of the requested item
+/// \return
+///
+
+///
+/// \fn requestPepItem(const QString &jid, const QString &nodeName, StandardItemId itemId = Current)
```
melvo, line 829:
```suggestion
/// \fn QXmppPubSubManager::requestPepItem(const QString &jid, const QString &nodeName, StandardItemId itemId = Current)
```

src/client/QXmppPubSubManager.cpp
lnjX: please do it correctly with the enum in this commit already (instead of adding code that is removed in the next commit) :)

src/client/QXmppPubSubManager.cpp
```diff
@@ -163,6 +180,55 @@ QXmppPubSubManager::~QXmppPubSubManager()
 {
 }
 
+///
+/// Requests all features of a pubsub service via service discovery.
+///
+/// This uses a \xep{0030, Service Discovery} info request to get the service
+/// identities and features.
+///
+/// Only if the service is of type serviceType, its features are returned.
```
lnjX, line 189:
```suggestion
/// The features are only returned if the service is of type serviceType,
/// otherwise InvalidServiceType is returned.
```

src/client/QXmppPubSubManager.h
lnjX: I'm sorry if I change my mind on this, but I don't think that `requestItem(jid, node)` is clear if you're reading the code without additional knowledge about `requestItem`. (`requestCurrentItem(jid, node)` would make clear which item (but the name and the concept have other issues)).

If you think `QXmppPubSubManager::Current` is too long (which I can understand), you can of course use something like `using PubSub = QXmppPubSubManager;` and then `requestItem(jid, node, PubSub::Current)`.

I don't want to exclude that we add the default parameter later on. That's an option and absolutely no problem (but removing it later would be problematic).

melvo: I do not think that `QXmppPubSubManager::Current` is too long. But I think that methods with `StandardItemId` as a parameter could have had reasonable defaults. There are several locations where the default value could have been used by the upcoming OMEMO manager. If you make such changes in the future, please make sure to refactor it correctly and consistently.

PR #393, opened 2022-01-27
OMEMO data class improvements

melvo:
- [ ] Documentation:


src/base/QXmppOmemoData.cpp
```diff
@@ -786,3 +787,90 @@ bool QXmppOmemoElement::isOmemoElement(const QDomElement &element)
...
+/// \cond
+void QXmppOmemoIq::parseElementFromChild(const QDomElement &element)
+{
+    QDomElement child = element.firstChildElement();
+    d->omemoElement->parse(child);
+}
+
+void QXmppOmemoIq::toXmlElementFromChild(QXmlStreamWriter *writer) const
+{
+    d->omemoElement->toXml(writer);
```
lnjX, line 875: crashes if the optional does not contain a value

src/base/QXmppOmemoData.cpp
```diff
@@ -786,3 +787,90 @@ bool QXmppOmemoElement::isOmemoElement(const QDomElement &element)
...
+///
+/// \param other
+///
+QXmppOmemoIq &QXmppOmemoIq::operator=(const QXmppOmemoIq &other) = default;
+
+/// \cond
+void QXmppOmemoIq::parseElementFromChild(const QDomElement &element)
+{
+    QDomElement child = element.firstChildElement();
+    d->omemoElement->parse(child);
```
lnjX, line 870: crashes if the omemoElement hasn't been created before

src/base/QXmppOmemoIq.h
```diff
@@ -0,0 +1,38 @@
...
+{
+public:
+    QXmppOmemoIq();
+    QXmppOmemoIq(const QXmppOmemoIq &);
+    ~QXmppOmemoIq() override;
+
+    QXmppOmemoIq &operator=(const QXmppOmemoIq &other);
+
+    std::optional<QXmppOmemoElement> omemoElement();
+    void setOmemoElement(const std::optional<QXmppOmemoElement> omemoElement);
```
lnjX, line 25: What's the idea of the optional here?
Also, you might want to inherit from OmemoElement, so you can directly access the attributes from the IQ and there's no need for an additional QSharedDataPointer.

src/base/QXmppOmemoData.cpp
```diff
@@ -695,7 +695,7 @@ void QXmppOmemoElement::setPayload(const QByteArray &payload)
 ///
 /// \return the found OMEMO envelope
 ///
-std::optional<QXmppOmemoEnvelope> QXmppOmemoElement::searchEnvelope(const QString &recipientJid, uint32_t recipientDeviceId) const
+std::optional<QXmppOmemoEnvelope> QXmppOmemoElement::searchEnvelope(const QString &recipientJid, const uint32_t recipientDeviceId) const
```
lnjX, line 707: I know nobody else than you doing this, the general coding style in QXmpp is without this const when parameters are passed by value.

tests/util.h
```diff
@@ -58,6 +58,29 @@ static void serializePacket(T &packet, const QByteArray &xml)
...
+        qDebug() << "writing" << data;
+
+        if (data == processedXml) {
+            return;
+        }
+    }
+
+    QFAIL("No XML data equals the serialized packet");
+}
+
```
lnjX, line 83: This seems to be a very special case, I suggest you move it into the file where it's currently used. If this is later used in other places we can think about making it available to all tests.

Apart from that:
 * the success case is not logged (as QCOMPARE would do if it succeeds)
 * the packet is serialized `n` times
 * should use QVector or alternatively variadic template arguments

src/base/QXmppOmemoIq.h
melvo: The OMEMO element does not need to be optional here. I changed that. The upcoming OMEMO manager creates an OMEMO element and sets it within the `QXmppOmemoIq`. Thus, I think it is easier if `QXmppOmemoIq` does not inherit from `QXmppOmemoElement`. Furthermore, the IQ stanza **is** not an OMEMO element, it **contains** one.

tests/util.h
melvo: The success case is logged and the `QList` is a `QVector` now.

The new `serializePacket()` is handy in all cases where a data structure with an undefined order (such as `QHash` or `QMultiHash`) is serialized . So, I think it would be better to keep it in that file. But if you think that it will not occur in the future, I can move it to `tests/qxmppomemodata/tst_qxmppomemodata.cpp`.

> the packet is serialized n times

How could that be improved?

src/base/QXmppOmemoDeviceBundle.h
```diff
@@ -36,7 +36,8 @@ class QXMPP_EXPORT QXmppOmemoDeviceBundle
     void setSignedPublicPreKeySignature(const QByteArray &signature);
 
     QHash<uint32_t, QByteArray> publicPreKeys() const;
-    void setPublicPreKeys(const QMap<uint32_t, QByteArray> &keys);
+    void addPublicPreKey(uint32_t id, const QByteArray &key);
+    void removePublicPreKey(uint32_t id);
```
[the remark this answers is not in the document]
melvo: It is not used by the upcoming OMEMO manager because it removes the whole bundle on a reset. Do you have a use case?

tests/util.h
lnjX: Testing it this way is a bad approach as it doesn't scale: you need to check against n! possible permutations, so for `n >= 3` you'll probably need a code generator for all permutations. Not to talk about the case when there's a randomly ordered list of elements with randomly ordered subelements.
That's why I don't want to make it 'official' by putting it in the util.h header file.

> > the packet is serialized n times
>
> How could that be improved?

do `const auto data = packetToXml(packet);` once, do the comparison in the loop.

src/base/QXmppOmemoDeviceBundle.h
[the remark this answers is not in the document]
lnjX: Currently if you want to remove all prekeys, you'd have to call `removePublicPreKey()` in a loop with all values from `publicPreKeys()`. If this class is shipped as part of the public API it'd be good to have such a function.

If the class isn't meant to be used by anyone else than the OmemoManager, it's maybe a good idea to make the class private within QXmpp.

tests/util.h
melvo: Thanks. I moved it to `tst_qxmppomemodata.cpp` and the serialization is only done once now.

src/base/QXmppOmemoIq.h
lnjX: > The upcoming OMEMO manager creates an OMEMO element and sets it within the QXmppOmemoIq. Thus, I think it is easier if QXmppOmemoIq does not inherit from QXmppOmemoElement.

Should be no big issue, you could e.g. add a `QXmppOmemoElement` constructor.

> Furthermore, the IQ stanza is not an OMEMO element, it contains one.

That's actually not the case, if you have a look at the XML. The IQ only consists of the omemo element, it's not like this: `<iq><omemo-iq><omemo-element/></omemo-iq></iq>`, but `<iq><omemo-element/></iq>` (and IQs have exactly one child element, so it wouldn't be possible to have any other extension).

src/base/QXmppOmemoData.cpp
```diff
@@ -786,3 +787,90 @@ bool QXmppOmemoElement::isOmemoElement(const QDomElement &element)
...
+///
+/// Determines whether the given DOM element is an OMEMO IQ stanza.
+///
+/// \param element DOM element being checked
+///
+/// \return true if element is an OMEMO IQ stanza, otherwise false
+///
+bool QXmppOmemoIq::isOmemoIq(const QDomElement &element)
+{
+    const QDomElement &child = element.firstChildElement();
```
lnjX, line 887: `&` is wrong here, firstChildElement() returns `QDomElement` (no ref)

```suggestion
    auto child = element.firstChildElement();
```

src/base/QXmppOmemoDeviceBundle.h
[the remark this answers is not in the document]
melvo: See 666aa730cd217cc6354351ed95f59dc947a0140e.

src/base/QXmppOmemoData.cpp
melvo: See 75d27e8093d900044c305a0cdfbed3e32784028c

src/base/QXmppOmemoData.cpp
```diff
@@ -24,20 +25,10 @@
 /// \since QXmpp 1.5
```
lnjX, line 25: it's probably a good idea to remove the `\since QXmpp 1.5` since it's not part of the public API

lnjX:
The build still fails in tst_qxmppmessage:
```
tst_qxmppmessage.obj : error LNK2019: unresolved external symbol "public: __cdecl QXmppOmemoElement::QXmppOmemoElement(void)" (??0QXmppOmemoElement@@QEAA@XZ) referenced in function "private: void __cdecl tst_QXmppMessage::testOmemoElement(void)" (?testOmemoElement@tst_QXmppMessage@@AEAAXXZ) [C:\projects\qxmpp\build\tests\tst_qxmppmessage.vcxproj]
```

I'd like to avoid making tst_qxmppmessage "internal" (because it's disabled by default then), so maybe you could move the omemoelement related tests into tst_qxmppomemodata?

melvo:
> The build still fails in tst_qxmppmessage:
> 
> ```
> tst_qxmppmessage.obj : error LNK2019: unresolved external symbol "public: __cdecl QXmppOmemoElement::QXmppOmemoElement(void)" (??0QXmppOmemoElement@@QEAA@XZ) referenced in function "private: void __cdecl tst_QXmppMessage::testOmemoElement(void)" (?testOmemoElement@tst_QXmppMessage@@AEAAXXZ) [C:\projects\qxmpp\build\tests\tst_qxmppmessage.vcxproj]
> ```
> 
> I'd like to avoid making tst_qxmppmessage "internal" (because it's disabled by default then), so maybe you could move the omemoelement related tests into tst_qxmppomemodata?

It is fixed now.

lnjX (approved):

PR #394, opened 2022-03-01
Use upper camel case for encryption method names

melvo:
- [ ] Documentation:


lnjX:
I don't see a big problem with the current capitalization (that is keeping uppercase letters uppercase). Your proposed version is probably more common, but changing this would require adding aliases, deprecating the old enums, documenting both variants  (with the new having \since 1.5) and possible users need to migrate and I don't think it's really worth it.

PR #396, opened 2022-03-05
Message: Add end-to-end encryption fallback body

melvo:
- [ ] Documentation:

### DIFF
--- a/src/base/QXmppMessage.cpp
+++ b/src/base/QXmppMessage.cpp
@@ -88,6 +88,7 @@ public:
     QXmppMessagePrivate();
 
     QString body;
+    QString e2eeFallbackBody;
     QString subject;
     QString thread;
     QString parentThread;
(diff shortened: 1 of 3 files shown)


src/base/QXmppMessage.cpp
```diff
@@ -236,6 +237,32 @@ void QXmppMessage::setBody(const QString &body)
...
+/// \since QXmpp 1.5
+///
+QString QXmppMessage::e2eeFallbackBody() const
+{
+    return d->e2eeFallbackBody;
+}
+
+///
+/// Sets an unencrypted string used to be presented to users if the message
+/// could not be decrypted.
```
lnjX, line 255: The QStrings are always unencrypted. How about this?

Sets an additional message body that is unlike the normal body not encrypted. This way you can have a fallback body for users not supporting the used end-to-end-encryption.

src/base/QXmppMessage.cpp
```diff
@@ -1442,7 +1474,22 @@ bool QXmppMessage::parseExtension(const QDomElement &element, QXmpp::SceMode sce
...
+                writer->writeDefaultNamespace(baseNamespace);
+            }
+            writer->writeCharacters(text);
+            writer->writeEndElement();
+        }
+    };
+
     if (sceMode & QXmpp::ScePublic) {
+        if (sceMode == QXmpp::ScePublic) {
+            writeTextElement(QStringLiteral("body"), d->e2eeFallbackBody);
```
lnjX, line 1490: writeTextElement() is only for private elements, use helperToXmlAddTextElement()

melvo: I used `writeTextElement()` instead because it should not be an empty element if the fallback body is empty.

src/base/QXmppMessage.cpp
```diff
@@ -1443,6 +1479,10 @@ bool QXmppMessage::parseExtension(const QDomElement &element, QXmpp::SceMode sce
 void QXmppMessage::serializeExtensions(QXmlStreamWriter *writer, QXmpp::SceMode sceMode, const QString &baseNamespace) const
 {
     if (sceMode & QXmpp::ScePublic) {
+        if (sceMode == QXmpp::ScePublic && !d->e2eeFallbackBody.isEmpty()) {
+            writer->writeTextElement(QStringLiteral("body"), d->e2eeFallbackBody);
+        }
```
lnjX, line 1484: As I said [earlier](https://github.com/qxmpp-project/qxmpp/pull/396#discussion_r827740415) please use  `helperToXmlAddTextElement()`

lnjX (approved):

PR #404, opened 2022-03-12
Fix some unsafety issues

JBBgameich:
- [x] Documentation:

### DIFF
--- a/src/client/QXmppUploadRequestManager.cpp
+++ b/src/client/QXmppUploadRequestManager.cpp
@@ -220,7 +220,8 @@ void QXmppUploadRequestManager::handleDiscoInfo(const QXmppDiscoveryIq &iq)
 
             // get size limit
             bool isFormNsCorrect = false;
-            for (const QXmppDataForm::Field &field : iq.form().fields()) {
+            const auto fields = iq.form().fields();
+            for (const QXmppDataForm::Field &field : fields) {
                 if (field.key() == QStringLiteral("FORM_TYPE")) {
                     isFormNsCorrect = field.value() == ns_http_upload;
                 } else if (isFormNsCorrect && field.key() == QStringLiteral("max-file-size")) {


src/client/QXmppUploadRequestManager.cpp
```diff
@@ -220,7 +220,8 @@ void QXmppUploadRequestManager::handleDiscoInfo(const QXmppDiscoveryIq &iq)
 
             // get size limit
             bool isFormNsCorrect = false;
-            for (const QXmppDataForm::Field &field : iq.form().fields()) {
+            const auto fields = iq.form().fields();
+            for (const QXmppDataForm::Field &field : fields) {
```
0xd34df00d, line 224: This is a good catch. So, here `iq.form()` returns a copy of the internal field, which would be temporary living till the end of the expression and whose lifetime will not be extended. `fields()`, though, will choose the non-`const` overload, returning the reference to the field of the temporary, whose lifetime will not be extended as well. Am I correct in following this?

src/client/QXmppTransferManager.cpp
```diff
@@ -1342,8 +1342,8 @@ void QXmppTransferManager::streamInitiationResultReceived(const QXmppStreamIniti
         job->state() != QXmppTransferJob::OfferState)
         return;
 
-    const auto &form = iq.featureForm();
-    const auto &fields = form.fields();
+    const auto form = iq.featureForm();
+    const auto fields = form.fields();
```
0xd34df00d, line 1346: I'm not sure why this is dangerous. `featureForm()` returns by value, so `const auto &form` would extend the lifetime, and so would `const auto &fields` for the return value of `fields()` (which would be by-value, since the `const` overload is chosen).

src/client/QXmppUploadRequestManager.cpp
JBBgameich: I think we are hitting https://github.com/KDE/clazy/blob/master/docs/checks/README-detaching-temporary.md here.
I was initially looking for what causes this asan error, and I think it's likely this line.
```
SUMMARY: AddressSanitizer: heap-use-after-free /usr/include/x86_64-linux-gnu/qt5/QtCore/qlist.h:178:32 in QList<QXmppDataForm::Field>::detach()
```

src/client/QXmppTransferManager.cpp
JBBgameich: To be honest I'm not completely sure anymore if this is a problem, maybe you know more on this than I do.
If the lifetime is extended by taking the ref, it should be fine I think.

0xd34df00d (approved):

PR #406, opened 2022-03-13
Fix various clazy warnings

lnjX:


src/base/QXmppJingleIq.cpp
```diff
@@ -461,8 +461,7 @@ void QXmppJingleIq::Content::toXml(QXmlStreamWriter *writer) const
 bool QXmppJingleIq::Content::parseSdp(const QString &sdp)
 {
     QList<QXmppJinglePayloadType> payloads;
-    QString line;
-    for (auto &line : sdp.split('\n')) {
+    for (auto &line : sdp.split(QChar(u'\n'))) {
```
JBBgameich, line 464: Isn't this also iterating over a temporary?

src/base/QXmppMessage.cpp
```diff
@@ -1547,7 +1547,7 @@ void QXmppMessage::serializeExtensions(QXmlStreamWriter *writer, QXmpp::SceMode
             writer->writeDefaultNamespace(ns_xhtml_im);
             writer->writeStartElement(QStringLiteral("body"));
             writer->writeDefaultNamespace(ns_xhtml);
-            writer->writeCharacters(QStringLiteral(""));
+            writer->writeCharacters(QLatin1String(""));
```
0xd34df00d, line 1550: Hmm, could this line be removed completely? I mean, what's the effect of `writeCharacters` with an empty string anyway?

src/base/QXmppJingleIq.cpp
```diff
@@ -497,7 +496,9 @@ bool QXmppJingleIq::Content::parseSdp(const QString &sdp)
                         if (payload.name() == QStringLiteral("telephone-event")) {
                             params.insert(QStringLiteral("events"), paramStr);
                         } else {
-                            for (const auto &p : paramStr.split(QRegularExpression(";\\s*"))) {
+                            thread_local static const auto regex = QRegularExpression(";\\s*");
```
0xd34df00d, line 499: Does it need to be `thread_local`?

0xd34df00d (approved):
Good!

As a nice side effect, quite a few cases of the "iterate over `keys()` of an associative container and grab the corresponding `value`" anti-pattern were removed, improving CPU and memory performance.

src/base/QXmppJingleIq.cpp
0xd34df00d: This shall be fine, the temporary will live long enough, and the container shouldn't be detached since its refcount will be `1`.

src/base/QXmppJingleIq.cpp
lnjX: QRegularExpression at least doesn't say it's thread-safe and from what I read (https://stackoverflow.com/questions/25581696/qregularexpression-matching-thread-safety ) that's probably required.

0xd34df00d: Interesting.

Let's play it safe here then, the overhead of `thread_local` is insignificant compared to the rest of what the code does, let alone the networking involved here.

src/base/QXmppMessage.cpp
lnjX: Unfortunately not because the underlying device() is used directly below; the tests fail without it:
```
FAIL!  : tst_QXmppMessage::testXhtml() Compared values are not the same
   Actual   (data)        : "<message type=\"normal\"><body>hi!</body><html xmlns=\"http://jabber.org/protocol/xhtml-im\"><body xmlns=\"http://www.w3.org/1999/xhtml\"<p style=\"font-weight:bold\">hi!</p>/></html></message>"
   Expected (processedXml): "<message type=\"normal\"><body>hi!</body><html xmlns=\"http://jabber.org/protocol/xhtml-im\"><body xmlns=\"http://www.w3.org/1999/xhtml\"><p style=\"font-weight:bold\">hi!</p></body></html></message>"
   Loc: [/home/lnj/dev/qxmpp/tests/./util.h(58)]
```

lnjX: Ah wait but what does work is writing a null string.

0xd34df00d: Ah, I see, so it's here for the writer to finish writing the opening `body` tag. Welp, this is so fragile, but I guess there are no better options.

PR #409, opened 2022-03-17
AtmManager: Fix memory leaks in tests and apply clang-format

melvo:
- [ ] Documentation:

### DIFF
--- a/tests/qxmppatmmanager/tst_qxmppatmmanager.cpp
+++ b/tests/qxmppatmmanager/tst_qxmppatmmanager.cpp
@@ -61,14 +61,13 @@ private:
     QXmppClient m_client;
     QXmppLogger m_logger;
     QXmppAtmTrustMemoryStorage m_trustStorage;
-    QXmppAtmManager *m_manager;
+    QXmppAtmManager m_manager { &m_trustStorage };
     QXmppCarbonManager *m_carbonManager;
 };
 
 void tst_QXmppAtmManager::initTestCase()
 {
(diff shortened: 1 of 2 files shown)


tests/qxmppatmmanager/tst_qxmppatmmanager.cpp
```diff
@@ -60,15 +60,14 @@ private slots:
...
     QXmppAtmManager *m_manager;
-    QXmppAtmTrustMemoryStorage *m_trustStorage;
     QXmppCarbonManager *m_carbonManager;
 };
 
 void tst_QXmppAtmManager::initTestCase()
 {
-    m_trustStorage = new QXmppAtmTrustMemoryStorage;
-    m_manager = new QXmppAtmManager(m_trustStorage);
+    m_manager = new QXmppAtmManager(&m_trustStorage);
```
0xd34df00d, line 70: Since we're changing this anyway, does it need to be heap-allocated? That is, why not have
```cpp
QXmppAtmManager m_manager { &m_trustStorage };
```
initializer in the class definition?

0xd34df00d (approved):
This already is a useful improvement, but I'm curious if `m_manager`'s lifetime could be made more explicit.

tests/qxmppatmmanager/tst_qxmppatmmanager.cpp
melvo: Thanks, I changed it accordingly.

PR #412, opened 2022-04-01
Move build constants into separate file

melvo:
- [ ] Documentation:


lnjX:
You can't change the contents that QXmppGlobal.h includes. The stuff from the new BuildConstants is missing.

I'd also suggest to only put the possible minimum into BuildConstants, so only the lines where `@...@` is used.

lnjX:
Sorry, the "closed" was a mistake (I merged it to master before pushing it here).

PR #414, opened 2022-04-04
ClientExtension: Add handleStanza() overload with e2ee metdata

lnjX:


lnjX:
Most of this has been merged already, other parts might follow later on.

PR #425, opened 2022-05-09
Message: Do not allow XEP-0184 request and received element together

melvo:
…not written

- [ ] Documentation:


lnjX:
If you've seen this behaviour, it would be good to fix it in the manager.

PR #431, opened 2022-05-25
cmake: add QXmppExtension.h to INSTALL_HEADER_FILES

tehnick:
This header is used by program Kaidan at least.


lnjX (approved):

lnjX:
Thanks for fixing this

PR #432, opened 2022-05-25
QXmppCallStream: use new GStreamer method

tehnick:
instead of deprecated one  when build with GStreamer >= 1.20.0


lnjX (approved):
Thanks

PR #434, opened 2022-05-30
Find dependencies dependencies of QXmpp when finding QXmpp

JBBgameich:
This way, users of the library don't need to find QtNetwork and QtXml themselves before finding QXmpp

PR check list:
- [x] Document your code
- [x] Add `\since QXmpp 1.X`
- [x] Update `doc/xep.doc`
- [x] Add unit tests


lnjX (approved):
Tested it, works well, thanks.

PR #439, opened 2022-08-13
Skip all integration tests if disabled

melvo:
PR check list:
- [x] Document your code
- [x] Add `\since QXmpp 1.X`
- [x] Update `doc/xep.doc`
- [x] Add unit tests

### DIFF
--- a/tests/qxmppiceconnection/tst_qxmppiceconnection.cpp
+++ b/tests/qxmppiceconnection/tst_qxmppiceconnection.cpp
@@ -4,6 +4,7 @@
 
 #include "QXmppStun.h"
 
+#include "IntegrationTesting.h"
 #include "util.h"
 #include <QHostInfo>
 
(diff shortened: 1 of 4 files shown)


lnjX:
That's no integration test. I know that these tests can cause problems, but marking them as "integration tests" is questionable.

tests/qxmppiceconnection/tst_qxmppiceconnection.cpp
```diff
@@ -19,6 +20,8 @@ private slots:
 
 void tst_QXmppIceConnection::testBind()
 {
+    SKIP_IF_INTEGRATION_TESTS_DISABLED();
```
lnjX, line 23: just a note: `SKIP_IF_INTEGRATION_TESTS_DISABLED()` doesn't need an extra `;` at the end

melvo:
> That's no integration test. I know that these tests can cause problems, but marking them as "integration tests" is questionable.

Thanks. How could that problem be solved?

lnjX:
> Thanks. How could that problem be solved?

Fixing the tests, marking tests requiring a working internet connection as online tests, make online tests optional.

PR #455, opened 2022-09-12
Implement XEP-0293: Jingle RTP Feedback Negotiation stanza parts

melvo:
PR check list:
- [x] Document your code
- [x] Update `doc/doap.xml`
- [x] Add unit tests

### DIFF
--- a/src/base/QXmppJingleIq.cpp
+++ b/src/base/QXmppJingleIq.cpp
@@ -1939,12 +1939,6 @@ bool QXmppJingleRtpFeedbackTypeAndParameters::isJingleRtpFeedbackTypeAndParamete
         element.namespaceURI() == ns_jingle_rtp_feedback_negotiation;
 }
 
-class QXmppJingleRtpFeedbackIntervalPrivate : public QSharedData
-{
-public:
-    uint32_t value;
-};
-
 ///
 /// \class QXmppJingleRtpFeedbackInterval
 ///
@@ -1958,7 +1952,6 @@ public:
 /// Constructs a Jingle RTP Feedback Negotiation interval.
 ///
 QXmppJingleRtpFeedbackInterval::QXmppJingleRtpFeedbackInterval()
-    : d(new QXmppJingleRtpFeedbackIntervalPrivate())
 {
 }
 
@@ -1969,9 +1962,9 @@ QXMPP_PRIVATE_DEFINE_RULE_OF_SIX(QXmppJingleRtpFeedbackInterval)
 ///
 /// \return the RTP Feedback Negotiation interval value
 ///
-uint32_t QXmppJingleRtpFeedbackInterval::value() const
+uint64_t QXmppJingleRtpFeedbackInterval::value() const
 {
-    return d->value;
+    return m_value;
 }
 
 ///
@@ -1979,22 +1972,22 @@ uint32_t QXmppJingleRtpFeedbackInterval::value() const
 ///
 /// \param value RTP Feedback Negotiation interval value
 ///
-void QXmppJingleRtpFeedbackInterval::setValue(uint32_t value)
+void QXmppJingleRtpFeedbackInterval::setValue(uint64_t value)
 {
-    d->value = value;
+    m_value = value;
 }
 
 /// \cond
 void QXmppJingleRtpFeedbackInterval::parse(const QDomElement &element)
 {
-    d->value = element.attribute(QStringLiteral("value")).toUInt();
+    m_value = element.attribute(QStringLiteral("value")).toUInt();
 }
 
 void QXmppJingleRtpFeedbackInterval::toXml(QXmlStreamWriter *writer) const
 {
     writer->writeStartElement(QStringLiteral("rtcp-fb-trr-int"));
     writer->writeDefaultNamespace(ns_jingle_rtp_feedback_negotiation);
-    helperToXmlAddAttribute(writer, QStringLiteral("value"), QString::number(d->value));
+    helperToXmlAddAttribute(writer, QStringLiteral("value"), QString::number(m_value));
     writer->writeEndElement();
 }
 /// \endcond
--- a/src/base/QXmppJingleIq.cpp
+++ b/src/base/QXmppJingleIq.cpp
@@ -146,6 +146,10 @@ public:
 
     QList<QXmppJinglePayloadType> payloadTypes;
     QList<QXmppJingleCandidate> transportCandidates;
+
+    // XEP-0293: Jingle RTP Feedback Negotiation
+    QVector<QXmppJingleRtpFeedbackTypeAndParameters> rtpFeedbackTypesAndParameters;
+    QVector<QXmppJingleRtpFeedbackInterval> rtpFeedbackIntervals;
 };
 
 QXmppJingleIqContentPrivate::QXmppJingleIqContentPrivate()
@@ -319,6 +323,50 @@ void QXmppJingleIq::Content::setTransportPassword(const QString &password)
 }
 
 ///
+/// Returns the types and parameters for RTP Feedback Negotiation.
+///
+/// \return the RTP Feedback Negotiation types and parameters
+///
+/// \since QXmpp 1.5
+///
+QVector<QXmppJingleRtpFeedbackTypeAndParameters> QXmppJingleIq::Content::rtpFeedbackTypesAndParameters() const
+{
+    return d->rtpFeedbackTypesAndParameters;
+}
+
+///
+/// Sets the types and parameters for RTP Feedback Negotiation.
+///
+/// \param type RTP Feedback Negotiation types and parameters
+///
+/// \since QXmpp 1.5
+///
+void QXmppJingleIq::Content::setRtpFeedbackTypesAndParameters(const QVector<QXmppJingleRtpFeedbackTypeAndParameters> &rtpFeedbackTypesAndParameters)
+{
+    d->rtpFeedbackTypesAndParameters = rtpFeedbackTypesAndParameters;
+}
+
+///
+/// Returns the intervals for RTP Feedback Negotiation.
+///
+/// \return the RTP Feedback Negotiation intervals
+///
+QVector<QXmppJingleRtpFeedbackInterval> QXmppJingleIq::Content::rtpFeedbackIntervals() const
+{
+    return d->rtpFeedbackIntervals;
+}
+
+///
+/// Sets the intervals for RTP Feedback Negotiation.
+///
+/// \param type RTP Feedback Negotiation intervals
+///
+void QXmppJingleIq::Content::setRtpFeedbackIntervals(const QVector<QXmppJingleRtpFeedbackInterval> &rtpFeedbackIntervals)
+{
+    d->rtpFeedbackIntervals = rtpFeedbackIntervals;
+}
+
+///
 /// Returns the fingerprint hash value for the transport key.
 ///
 /// This is used for DTLS-SRTP as defined in \xep{0320}.
@@ -424,9 +472,11 @@ void QXmppJingleIq::Content::parse(const QDomElement &element)
         d->transportCandidates << candidate;
         child = child.nextSiblingElement(QStringLiteral("candidate"));
     }
-    child = transportElement.firstChildElement(QStringLiteral("fingerprint"));
+
+    parseJingleRtpFeedbackNegotiationElements(descriptionElement, d->rtpFeedbackTypesAndParameters, d->rtpFeedbackIntervals);
 
     /// XEP-0320
+    child = transportElement.firstChildElement(QStringLiteral("fingerprint"));
     if (!child.isNull()) {
         d->transportFingerprint = parseFingerprint(child.text());
         d->transportFingerprintHash = child.attribute(QStringLiteral("hash"));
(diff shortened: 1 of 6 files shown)


src/base/QXmppJingleIq.cpp
```diff
@@ -1462,3 +1572,366 @@ bool QXmppJinglePayloadType::operator==(const QXmppJinglePayloadType &other) con
...
+{
+    return element.tagName() == QStringLiteral("rtcp-fb") &&
+        element.namespaceURI() == ns_jingle_rtp_feedback_negotiation;
+}
+
+class QXmppJingleRtpFeedbackIntervalPrivate : public QSharedData
+{
+public:
+    uint32_t value;
+};
```
lnjX, line 1826: this is so small, it shouldn't have a d-ptr and please use uint64_t (so we later have the option to add a d ptr)

src/base/QXmppJingleIq.h
```diff
@@ -317,4 +407,7 @@ class QXMPP_EXPORT QXmppJingleIq : public QXmppIq
     QSharedDataPointer<QXmppJingleIqPrivate> d;
 };
 
+static void parseJingleRtpFeedbackNegotiationElements(const QDomElement &parent, QVector<QXmppJingleRtpFeedbackTypeAndParameters> &typesAndParameters, QVector<QXmppJingleRtpFeedbackInterval> &intervals);
+static void jingleRtpFeedbackNegotiationElementsToXml(QXmlStreamWriter *writer, const QVector<QXmppJingleRtpFeedbackTypeAndParameters> &typesAndParameters, const QVector<QXmppJingleRtpFeedbackInterval> &intervals);
```
lnjX, line 411: why is this public? and in global scope without QXmpp in the name?

src/base/QXmppJingleIq.h
```diff
@@ -13,6 +14,83 @@ class QXmppJingleCandidatePrivate;
 class QXmppJingleIqContentPrivate;
 class QXmppJingleIqPrivate;
 class QXmppJinglePayloadTypePrivate;
+class QXmppJingleRtpFeedbackIntervalPrivate;
+class QXmppJingleRtpFeedbackTypeAndParametersPrivate;
+class QXmppSdpParameterPrivate;
+
+// TODO: Use QXMPP_PRIVATE_DECLARE_RULE_OF_SIX and QXMPP_PRIVATE_DEFINE_ROLE_OF_SIX
```
lnjX, line 21: .

src/base/QXmppJingleIq.h
```diff
@@ -13,6 +14,83 @@ class QXmppJingleCandidatePrivate;
...
+    void toXml(QXmlStreamWriter *writer) const;
+    /// \endcond
+
+    static bool isSdpParameter(const QDomElement &element);
+
+private:
+    QSharedDataPointer<QXmppSdpParameterPrivate> d;
+};
+
+class QXMPP_EXPORT QXmppJingleRtpFeedbackTypeAndParameters
```
lnjX, line 47: Can we have a better name here (without `and` in it)?

src/base/QXmppJingleIq.cpp
```diff
@@ -1462,3 +1572,366 @@ bool QXmppJinglePayloadType::operator==(const QXmppJinglePayloadType &other) con
...
+
+/// \cond
+///
+/// Parses the RTP Feedback Negotation elements.
+///
+/// \param parent element to parse which might contain RTP Feedback Negotiation elements
+/// \param typesAndParameters RTP Feedback Negotation types and parameters to be set
+/// \param intervals RTP Feedback Negotiation intervals to be set
+///
+void parseJingleRtpFeedbackNegotiationElements(const QDomElement &parent, QVector<QXmppJingleRtpFeedbackTypeAndParameters> &typesAndParameters, QVector<QXmppJingleRtpFeedbackInterval> &intervals)
```
lnjX, line 1903: can `static` be omitted here? there are no complaints if static is present.

src/base/QXmppJingleIq.cpp
```diff
@@ -1462,3 +1572,366 @@ bool QXmppJinglePayloadType::operator==(const QXmppJinglePayloadType &other) con
...
+}
+
+///
+/// Serializes the RTP Feedback Negotation elements.
+///
+/// \param writer writer to write the serialized elements
+/// \param typesAndParameters RTP Feedback Negotation types and parameters to be serialized
+/// \param intervals RTP Feedback Negotiation intervals to be serialized
+///
+void jingleRtpFeedbackNegotiationElementsToXml(QXmlStreamWriter *writer, const QVector<QXmppJingleRtpFeedbackTypeAndParameters> &typesAndParameters, const QVector<QXmppJingleRtpFeedbackInterval> &intervals)
```
lnjX, line 1927: same note about `static`

lnjX (changes requested):

src/base/QXmppJingleIq.cpp
```diff
@@ -1462,3 +1572,366 @@ bool QXmppJinglePayloadType::operator==(const QXmppJinglePayloadType &other) con
...
+
+void QXmppJingleRtpFeedbackTypeAndParameters::toXml(QXmlStreamWriter *writer) const
+{
+    writer->writeStartElement(QStringLiteral("rtcp-fb"));
+    writer->writeDefaultNamespace(ns_jingle_rtp_feedback_negotiation);
+    helperToXmlAddAttribute(writer, QStringLiteral("type"), d->type);
+
+    // If there are parameters, they must be used instead of the subtype.
+    if (const auto subtype = d->subtype; subtype.isEmpty()) {
+        for (const auto &parameter : std::as_const(d->parameters)) {
```
lnjX, line 1797: `d->parameters` is already const here, because `toXml()` is marked as const.

src/base/QXmppJingleIq.cpp
```diff
@@ -288,6 +293,50 @@ void QXmppJingleIq::Content::setTransportPassword(const QString &password)
...
+
+///
+/// Sets the intervals for RTP Feedback Negotiation.
+///
+/// \param type RTP Feedback Negotiation intervals
+///
+void QXmppJingleIq::Content::setRtpFeedbackIntervals(const QVector<QXmppJingleRtpFeedbackInterval> &rtpFeedbackIntervals)
+{
+    d->rtpFeedbackIntervals = rtpFeedbackIntervals;
+}
```
lnjX, line 338: can you test this by creating a new element using the setters and then serializing it?

src/base/QXmppJingleIq.cpp
```diff
@@ -1462,3 +1616,316 @@ bool QXmppJinglePayloadType::operator==(const QXmppJinglePayloadType &other) con
...
+
+void QXmppJingleRtpFeedbackProperty::toXml(QXmlStreamWriter *writer) const
+{
+    writer->writeStartElement(QStringLiteral("rtcp-fb"));
+    writer->writeDefaultNamespace(ns_jingle_rtp_feedback_negotiation);
+    helperToXmlAddAttribute(writer, QStringLiteral("type"), d->type);
+
+    // If there are parameters, they must be used instead of the subtype.
+    if (const auto subtype = d->subtype; subtype.isEmpty()) {
```
lnjX, line 1842: why are you copying d->subtype??

src/base/QXmppJingleIq.cpp
```diff
@@ -1462,3 +1616,316 @@ bool QXmppJinglePayloadType::operator==(const QXmppJinglePayloadType &other) con
...
+///
+/// Determines whether the given DOM element is a Session Description Protocol parameter element.
+///
+/// \param element DOM element being checked
+///
+/// \return whether element is a Session Description Protocol parameter element
+///
+bool QXmppSdpParameter::isSdpParameter(const QDomElement &element)
+{
+    return element.tagName() == QStringLiteral("parameter");
```
lnjX, line 1719: missing namespace check

src/base/QXmppJingleIq.cpp
```diff
@@ -1462,3 +1616,316 @@ bool QXmppJinglePayloadType::operator==(const QXmppJinglePayloadType &other) con
...
+    d->name = element.attribute(QStringLiteral("name"));
+    d->value = element.attribute(QStringLiteral("value"));
+}
+
+void QXmppSdpParameter::toXml(QXmlStreamWriter *writer) const
+{
+    writer->writeStartElement(QStringLiteral("parameter"));
+    helperToXmlAddAttribute(writer, QStringLiteral("name"), d->name);
+
+    if (const auto value = d->value; !value.isEmpty()) {
```
lnjX, line 1702: don't copy d->value

src/base/QXmppJingleIq.cpp
```diff
@@ -120,6 +121,50 @@ static QString candidateToSdp(const QXmppJingleCandidate &candidate)
...
 }
 
+/// \cond
+///
+/// Parses the RTP Feedback Negotation elements.
+///
+/// \param parent element to parse which might contain RTP Feedback Negotiation elements
+/// \param properties RTP feedback properties to be set
+/// \param intervals RTP feedback intervals to be set
+///
```
lnjX, line 131: why are you writing doxygen documentation for code that is private and excluded from the generated doxygen documentation?

src/base/QXmppJingleIq.h
```diff
@@ -316,5 +403,4 @@ class QXMPP_EXPORT QXmppJingleIq : public QXmppIq
 private:
     QSharedDataPointer<QXmppJingleIqPrivate> d;
 };
```
lnjX, line 405:
```suggestion
};

```

src/base/QXmppJingleIq.cpp
melvo: The parameter has no namespace. Thus, there is no check for it.

src/base/QXmppJingleIq.cpp
melvo: It is the way we use for documenting code. I added `\cond` and `\endcond` in order to exclude it from the generated (public) documentation. How should / could that be done otherwise?

lnjX: Okay, so here comes my analysis about such documentation:

Why do you think this documentation is helpful here? More documentation can also bring disadvantages, it could get outdated and can mean more maintenance effort. Which new information does the documentation provide here?

When reading this documentation
 * you are already seeing the whole function implementation, because we don't generate external documentation
 * you already know the function name and the parameters

The current doxygen documentation provides:
* `/// Parses the RTP Feedback Negotation elements.`
  The name of the function, without 'Jingle', without camel-case and with spaces.
* `/// \param parent element to parse which might contain RTP Feedback Negotiation elements`
  That this elment is the parent is already given by the name of the parameter, the only new information here is that it *might contain such elements*
* `/// \param properties RTP feedback properties to be set`
  `/// \param intervals RTP feedback intervals to be set`
  New information: the function writes to these referenced variables, which is also indicated by `&` without `const` (but not explicit)

To sum up, the new information is:
 1. `parent` is allowed to contain no valid elements
 1. it's now clearer that the function writes to `properties` and `intervals`

Have I missed something?

Our goals are:
 * to minimize maintenance efforts, so not repeating information and making it easy to update the documentation when changing the code
 * to provide concise and easy-to-read information about the function, so only providing what is actually helpful and again not repeating information
 * and to minimize the development time which can also be important

The significant information we have got here can be put into a normal comment.
It could look like this:
```cpp
// Parses all found property and interval elements into properties and intervals.
static void parseJingleRtpFeedbackNegotiationElements(const QDomElement &parent, QVector<QXmppJingleRtpFeedbackProperty> &properties, QVector<QXmppJingleRtpFeedbackInterval> &intervals)
{
    for (auto child = parent.firstChildElement();
         !child.isNull();
         child = child.nextSiblingElement()) {
...
    }
}
```

The alternative would be to add the comments directly at the lines where the action is going on. This has the advantage that the comments are less likely forgot when changing the function. However in this case the comments seem completely useless:
```cpp
static void parseJingleRtpFeedbackNegotiationElements(const QDomElement &parent, QVector<QXmppJingleRtpFeedbackProperty> &properties, QVector<QXmppJingleRtpFeedbackInterval> &intervals)
{
    // parent may contain no valid child elements
    for (auto child = parent.firstChildElement();
         !child.isNull();
         child = child.nextSiblingElement()) {
        // Write parsed elements to properties and intervals
        if (QXmppJingleRtpFeedbackProperty::isJingleRtpFeedbackProperty(child)) {
            QXmppJingleRtpFeedbackProperty property;
            property.parse(child);
            properties.append(property);
        } else if (QXmppJingleRtpFeedbackInterval::isJingleRtpFeedbackInterval(child)) {
            QXmppJingleRtpFeedbackInterval interval;
            interval.parse(child);
            intervals.append(interval);
        }
    }
}
```

Now the thing is
 * this function is very simple
 * it's really not hard to see both points from the comment in the code
 * if the behaviour of the function changes, the comments need to be adjusted
 * if someone changes the behaviour and forgets to update the comments, the code will contain comments that tell something misleading
 
Did I missing something? Are there advantages of the doxygen styled documentation with explicit `\param`s?

In this case I don't think that the overhead of the full doxygen documentation stays in a good relation to the advantages. A simple comment does the same job and the important information is not repeated. I will leave the decision whether such a comment is needed here at all open to you.

lnjX: That's a point. Doesn't qtcreator display hints for normal comments and does it display doxygen inside \cond? I personally don't really use the hints.

For simple functions or functions in the same file I'm still not really convinced.

lnjX (approved):

PR #467, opened 2022-09-21
Implement XEP-0167: Jingle RTP Sessions multiplexing

melvo:
PR check list:
- [x] Document your code
- [x] Update `doc/doap.xml`
- [x] Add unit tests

### DIFF
--- a/src/base/QXmppJingleIq.h
+++ b/src/base/QXmppJingleIq.h
@@ -188,6 +188,9 @@ public:
         quint32 descriptionSsrc() const;
         void setDescriptionSsrc(quint32 ssrc);
 
+        bool isRtpMultiplexingSupported() const;
+        void setRtpMultiplexingSupported(bool isRtpMultiplexingSupported);
+
         void addPayloadType(const QXmppJinglePayloadType &payload);
         QList<QXmppJinglePayloadType> payloadTypes() const;
         void setPayloadTypes(const QList<QXmppJinglePayloadType> &payloadTypes);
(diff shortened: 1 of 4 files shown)


tests/qxmppjingleiq/tst_qxmppjingleiq.cpp
```diff
@@ -59,72 +59,92 @@ void tst_QXmppJingleIq::testContent()
...
-        " ip=\"192.0.2.3\""
-        " network=\"1\""
-        " port=\"45664\""
-        " priority=\"1694498815\""
-        " protocol=\"udp\""
-        " type=\"srflx\"/>"
+        " network=\"0\""
+        " port=\"0\""
+        " priority=\"0\""
+        " type=\"host\"/>"
```
lnjX, line 82: why are you testing only with zeros (instead of realistic data)?

melvo: This tests only the parts of the content because the candidate is already tested by `testCandidate()`. By doing that, we do not recursively duplicate tests and instead focus only on what has to be tested here.

src/base/QXmppJingleIq.h
```diff
@@ -188,6 +188,9 @@ class QXMPP_EXPORT QXmppJingleIq : public QXmppIq
         quint32 descriptionSsrc() const;
         void setDescriptionSsrc(quint32 ssrc);
 
+        bool isRtpMultiplexingSupported() const;
+        void setIsRtpMultiplexingSupported(bool isRtpMultiplexingSupported);
```
lnjX, line 192:
```suggestion
        void setRtpMultiplexingSupported(bool isRtpMultiplexingSupported);
```

lnjX (approved):

PR #468, opened 2022-09-24
Rename OmemoEnvelope's 'setIsUsedForKeyExchange()' to 'setUsedForKeyExchange()'

melvo:
PR check list:
- [x] Document your code
- [x] Update `doc/doap.xml`
- [x] Add unit tests

### DIFF
--- a/src/base/QXmppOmemoDataBase.cpp
+++ b/src/base/QXmppOmemoDataBase.cpp
@@ -60,7 +60,7 @@ bool QXmppOmemoEnvelope::isUsedForKeyExchange() const
 ///
 /// \param isUsed whether a pre-key was used to prepare this envelope
 ///
-void QXmppOmemoEnvelope::setIsUsedForKeyExchange(bool isUsed)
+void QXmppOmemoEnvelope::setUsedForKeyExchange(bool isUsed)
 {
     m_isUsedForKeyExchange = isUsed;
 }
--- a/src/base/QXmppOmemoEnvelope_p.h
+++ b/src/base/QXmppOmemoEnvelope_p.h
@@ -18,7 +18,7 @@ public:
     void setRecipientDeviceId(uint32_t id);
 
     bool isUsedForKeyExchange() const;
-    void setIsUsedForKeyExchange(bool isUsed);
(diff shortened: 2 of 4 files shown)


lnjX:
there are probably other places where this is the case, can you also adjust them?

lnjX (approved):

lnjX:
I'd like to stick to the Qt conventions as much as possible.

https://wiki.qt.io/API_Design_Principles#Naming_Boolean_Getters.2C_Setters.2C_and_Properties

It says:
* nouns shouldn't have `is` for the getter at all, except it would be misleading. (e.g. isMixChannel() instead of mixChannel())
* setters should never have `is` in it

At first I agreed that having a setter called `setMixChannel()` sounds weird, but the more I think about it, it sounds more and more reasonable to me. I think `setMixChannel(true)` is okay.

However, it's probably a good idea to use more explicit types such as enums or structs to avoid such situations. For example:
* `setSpoiler(bool)` (and `setSpoilerHint()`) could be replaced with `setSpoiler(optional<Spoiler>)` and a `struct Spoiler { QString hint; };`;
* `setMixChannel()` could be replaced with `setChatType(ChatType)` and an `enum ChatType { User, MixChannel }` (in this case this approach could lead to issues later tough, e.g. if the channel is also a MUC and MIX2 chat and that should also be represented)

I'd suggest to:
1. replace `setIsSpoiler()` with `setSpoiler(optional<Spoiler>)` + spoiler struct
2. `setIsFallback()`: wait with replacing until the contents of the new version of the fallback indicator are implemented and then use a fallback struct/class
3. remove `is` in `QXmppPresence::setIsPreparingMujiSession()`, `QXmppRegisterIq::setIsRegistered()`, `QXmppRegisterIq::setIsRemove()`, `QXmppRosterIq::Item::setIsApproved()`
4. replace `setIsMixChannel(bool)` with `setMixChannel(optional<MixChannel>)` and a struct/class `MixChannel { QString participantId; }`.

It would be great if you could do 3. in this PR, the other points can be done later (if you don't want to do them I can do them too). :)

PR #483, opened 2022-09-30
Implement XEP-0294: Jingle RTP Header Extensions Negotiation stanzas

melvo:
PR check list:
- [x] Document your code
- [x] Update `doc/doap.xml`
- [x] Add unit tests

### DIFF
--- a/src/base/QXmppJingleIq.cpp
+++ b/src/base/QXmppJingleIq.cpp
@@ -121,6 +121,27 @@ static QString candidateToSdp(const QXmppJingleCandidate &candidate)
     return QStringLiteral("candidate:%1 %2 %3 %4 %5 %6 typ %7 generation %8").arg(candidate.foundation(), QString::number(candidate.component()), candidate.protocol(), QString::number(candidate.priority()), candidate.host().toString(), QString::number(candidate.port()), QXmppJingleCandidate::typeToString(candidate.type()), QString::number(candidate.generation()));
 }
 
+// Parses all found SDP parameter elements of parent into parameters.
+static void parseSdpParameters(const QDomElement &parent, QVector<QXmppSdpParameter> &parameters)
+{
+    for (auto childElement = parent.firstChildElement();
+         !childElement.isNull();
+         childElement = childElement.nextSiblingElement()) {
+        if (QXmppSdpParameter::isSdpParameter(childElement)) {
+            QXmppSdpParameter parameter;
+            parameter.parse(childElement);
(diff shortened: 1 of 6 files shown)


src/base/QXmppJingleIq.cpp
```diff
@@ -2005,3 +2095,233 @@ bool QXmppJingleRtpFeedbackInterval::isJingleRtpFeedbackInterval(const QDomEleme
...
+}
+
+void QXmppJingleRtpHeaderExtensionProperty::toXml(QXmlStreamWriter *writer) const
+{
+    writer->writeStartElement(QStringLiteral("rtp-hdrext"));
+    writer->writeDefaultNamespace(ns_jingle_rtp_header_extensions_negotiation);
+    helperToXmlAddAttribute(writer, QStringLiteral("id"), QString::number(d->id));
+    helperToXmlAddAttribute(writer, QStringLiteral("uri"), d->uri);
+
+    if (const auto senders = d->senders; senders != QXmppJingleRtpHeaderExtensionProperty::Both) {
```
lnjX, line 2241: I'm actually interested, I've seen this now multiple times, what was the idea behind this construct?

src/base/QXmppJingleIq.h
```diff
@@ -89,6 +90,66 @@ class QXMPP_EXPORT QXmppJingleRtpFeedbackInterval
...
+    /// \cond
+    void parse(const QDomElement &element);
+    void toXml(QXmlStreamWriter *writer) const;
+    /// \endcond
+
+    static bool isJingleRtpHeaderExtensionMixing(const QDomElement &element);
+
+private:
+    bool m_isAllowed = false;
+};
```
lnjX, line 151: If this is really for just one bool, I think an extra class with parse(), toXml() etc. could be too much here? Do you think it is a good idea?

src/base/QXmppJingleIq.cpp
melvo: The intention was to use a variable instead of getting it multiple times. But as I see that you are in favor of the opposite, I will change it.

lnjX: You are copying it here. It's probably optimized away and it's an integer, so performance is not an argument, but I think the code just doesn't get easier to read.

lnjX (approved):

lnjX:
This broke the Qt 5.9 build. Could you fix it?

melvo:
Thanks. #489 should fix it.

PR #492, opened 2022-10-04
Implement XEP-0444: Message Reactions

melvo:
PR check list:
- [x] Document your code
- [x] Update `doc/doap.xml`
- [x] Add unit tests


melvo:
@lnjX I think that `reactions` would fit better than `reaction` because it is a list of emojis. What do you think?

lnjX:
5.9 and windows builds are broken

src/base/QXmppMessageReaction.h
```diff
@@ -0,0 +1,36 @@
+// SPDX-FileCopyrightText: 2022 Melvin Keskin <melvo@olomono.de>
+//
+// SPDX-License-Identifier: LGPL-2.1-or-later
+
+#ifndef QXMPPMESSAGEREACTION_H
+#define QXMPPMESSAGEREACTION_H
+
+#include "QXmppGlobal.h"
+
+#include "QList"
```
lnjX, line 10: ?

lnjX (changes requested):

lnjX:
> @lnjX I think that `reactions` would fit better than `reaction` because it is a list of emojis. What do you think?

yes, `reactions` would be better, it makes more sense, the XML element is also called reactions.

lnjX:
I think it's unfortunate that the XEP doesn't use message attaching (yes, I've read the note about it in the XEP), because you can't easily fetch all the reactions via MAM but you need to receive all later MAM reaction messages to display reactions correctly.

melvo:
> I think it's unfortunate that the XEP doesn't use message attaching (yes, I've read the note about it in the XEP), because you can't easily fetch all the reactions via MAM but you need to receive all later MAM reaction messages to display reactions correctly.
> 
> Edit: I actually meant Message Fastening (https://xmpp.org/extensions/xep-0422.html)

@mar-v-in What do you think about that?

src/base/QXmppMessageReaction.h
```diff
@@ -7,7 +7,7 @@
 
 #include "QXmppGlobal.h"
 
-#include "QList"
+#include "QVector"
```
lnjX: `<QVector>`

src/base/QXmppMessage.cpp
```diff
@@ -1255,14 +1255,16 @@ std::optional<QXmppMessageReaction> QXmppMessage::reaction() const
 ///
 /// Sets a reaction to a message as defined by \xep{0444, Message Reactions}.
 ///
-/// The corresponding storage hint is automatically set.
+/// The corresponding hint must be set manually:
+/// \code
+/// QXmppMessage::addHint(QXmppMessage::Store)
```
lnjX, line 1260: The function isn't static, so better use it on an object.

src/base/QXmppMessageReaction.cpp
```diff
@@ -7,24 +7,31 @@
...
 ///
 /// \class QXmppMessageReaction
 ///
-/// \brief The QXmppMessageReaction class represents a reaction to a message as specified by
-/// \xep{0444: Message Reactions} containing emojis.
-///
-/// Each element should only consist of unicode codepoints that can be displayed as a single emoji.
-/// Duplicates are not allowed.
+/// \brief The QXmppMessageReaction class represents a reaction to a message in the form of emojis
+/// as specified by \xep{0444: Message Reactions}.
```
lnjX, line 23: wrong \xep{} usage with `:`

src/base/QXmppMessageReaction.h
```diff
@@ -22,6 +23,9 @@ class QXMPP_EXPORT QXmppMessageReaction : public QVector<QString>
     QString id() const;
     void setId(const QString &id);
```
lnjX, line 24: Can you rename this to `messageId`? It's not the id of the reaction.

src/base/QXmppMessageReaction.cpp
```diff
@@ -0,0 +1,121 @@
...
+{
+}
+
+QXMPP_PRIVATE_DEFINE_RULE_OF_SIX(QXmppMessageReaction)
+
+///
+/// Returns the ID of the message for that the reaction is sent.
+///
+/// \return the message's ID
+///
```
lnjX, line 42: please explain which message id is meant here (there're the normal `id`, origin id, stanza id of the server, and maybe something i've missing)

lnjX (changes requested):

mar-v-in:
As of now, Fastening (XEP-0422) does not solve any issues here. MAM-FC (XEP-0427) - even if implemented by a server - would not be suitable for reactions as it doesn't preserve the information of the sender of the fastening in collate mode (simplified mode would only remove the reactions from MAM, full mode is what we have now). Sensible summarizing/collating of reactions would only be possible with server directly supporting reactions (and its processing rules), which means using fastening would not help realizing that.

lnjX:
@mar-v-in I see it's not ideal. I think what would still make sense is to fetch old messages in `collate` mode and to load all fastenings/all reaction messages with a `fastenings` query.

Currently I see a problem in the following scenario:
1. you join a group chat
2. you load the message history
3. you receive messages with reactions, but you can't find a message with the id => your client ignores the message
4. you load another old message which people have reacted to, but your client doesn't display them

There're (at least) two ways to workaround this a) to load all messages in chronological order or b) to store the reactions messages even if the message id is unknown when receiving the reaction.

I haven't thought about fastenings in detail, maybe message fastenings (or mam fc) still need some adjustments or changes in the architecture, but in general it sounds like the "right way" to do this.

mar-v-in:
- Dino stores the reactions even for unknown messages. I wouldn't consider this a workaround though.
- The problem is that you need to know who sent which reaction because you need to be able to process the live reactions when they come in (which means to *replace* what was there before from the same sender). Also you need to identity the user by occupant-id in anonymous MUCs.
- I disagree that fastening / MAM-FC is the right way to do it. It tries to generify over all possible things that can be attached to a message. I doubt this is possible. I'd rather think that it makes sense to include processing and summarizing rules for MAM in each XEP where it makes sense to have those.
- In practice, the summarized message isn't much smaller than the original message plus all its attachment, so the only advantage is that things are retrieved together. This indeed could be done in a generic way - one would probably want to have this even two-way (for replies, that would allow to instruct the server to also provide the message you replied to). I once drafted a specification for that, so if you think this is something that makes sense, we could build this in a XEP

src/base/QXmppMessageReaction.h
melvo: That is already removed.

src/base/QXmppMessage.cpp
melvo: It was meant as a reference to the method but I adjusted it to make it clear.

lnjX (approved):

lnjX:
@mar-v-in I see your point. My main goal would be to be able to receive all attached/related messages at once. Your draft sounds interesting and having it work both ways (also for replys) sounds very useful.

PR #507, opened 2022-10-26
OmemoManagerPrivate: Fix device setup

melvo:
PR check list:
- [x] Document your code
- [x] Update `doc/doap.xml`
- [x] Add unit tests

### DIFF
--- a/src/omemo/QXmppOmemoManager_p.cpp
+++ b/src/omemo/QXmppOmemoManager_p.cpp
@@ -616,10 +616,8 @@ signal_protocol_session_store ManagerPrivate::createSessionStore() const
 //
 QFuture<bool> ManagerPrivate::setUpDeviceId()
 {
-    QFutureInterface<bool> interface(QFutureInterfaceBase::Started);
-
     auto future = pubSubManager->requestPepItemIds(ns_omemo_2_bundles);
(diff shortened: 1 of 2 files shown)


melvo:
This change is especially needed in order to set up OMEMO with Prosody.

src/omemo/QXmppOmemoManager_p.cpp
```diff
@@ -620,24 +620,44 @@ QFuture<bool> ManagerPrivate::setUpDeviceId()
...
+                const auto &deviceIds = std::get<QVector<QString>>(result);
+
+                while (true) {
+                    uint32_t deviceId = 0;
+
+                    if (generateDeviceId(deviceId) && !deviceIds.contains(QString::number(deviceId))) {
+                        ownDevice.id = deviceId;
+                        reportFinishedResult(interface, true);
+                        break;
+                    }
```
lnjX, line 660: The error case / where `reportFinishedResult(interface, false);` is called is a bit hidden here and I don't think the helper function `generateDeviceId()` should be responsible for reporting the result of `setUpDeviceId()`.

lnjX (changes requested):

lnjX (approved):

PR #545, opened 2023-03-08
CarbonManagerV2: Fix enabling carbons

melvo:
PR check list:
- [x] Document your code
- [x] Update `doc/doap.xml`
- [x] Add unit tests


melvo:
@lnj The regular way for enabling carbons would be to not add a `to` element but I could not find a way to omit that. When I do not set the `to` element, it is automatically set to the server's JID.

lnjX (approved):
Thanks for the fix :)

PR #556, opened 2023-03-11
Stream: Allow no 'from' attribute for IQ responses

melvo:
See https://xmpp.org/rfcs/rfc6120.html#stanzas-attributes-from-c2s point 3

PR check list:
- [x] Document your code
- [x] Update `doc/doap.xml`
- [x] Add unit tests

### DIFF
--- a/src/base/QXmppStream.cpp
+++ b/src/base/QXmppStream.cpp
@@ -481,10 +481,14 @@ bool QXmppStream::handleIqResponse(const QDomElement &stanza)
         return false;
     }
 
-    if (auto itr = d->runningIqs.find(stanza.attribute(QStringLiteral("id")));
+    const auto id = stanza.attribute(QStringLiteral("id"));
+    if (auto itr = d->runningIqs.find(id);
         itr != d->runningIqs.end()) {
-        if (stanza.attribute("from") != itr.value().jid) {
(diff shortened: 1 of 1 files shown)


src/base/QXmppStream.cpp
```diff
@@ -481,10 +481,14 @@ bool QXmppStream::handleIqResponse(const QDomElement &stanza)
...
-    if (auto itr = d->runningIqs.find(stanza.attribute(QStringLiteral("id")));
+    const auto id = stanza.attribute(QStringLiteral("id"));
+    if (auto itr = d->runningIqs.find(id);
         itr != d->runningIqs.end()) {
-        if (stanza.attribute("from") != itr.value().jid) {
-            warning(QStringLiteral("Received IQ response to one of our requests from wrong sender. Ignoring."));
+        const auto expectedSenderJid = itr.value().jid;
+        // Stanzas coming from the server on behalf of the user's account must have no "from"
+        // attribute or have it set to the user's bare JID.
+        if (const auto senderJid = stanza.attribute("from"); !senderJid.isEmpty() && senderJid != expectedSenderJid) {
```
lnjX, line 490: what if the senderJid is empty and doesn't match the expected sender?

PR #575, opened 2023-06-21
XEP-0482: Call Invites

taiBsu:
PR check list:
- [x] Document your code
- [x] Update `doc/doap.xml`
- [x] Add unit tests


src/client/QXmppCallInviteManager.h
```diff
@@ -0,0 +1,123 @@
...
+    using Result = std::variant<Rejected, Retracted, Left, QXmppError>;
+
+    QXmppCallInvite(QXmppCallInviteManager *manager);
+    ~QXmppCallInvite();
+
+    QXmppTask<QXmpp::SendResult> invite(
+        bool audio = true,
+        bool video = false,
+        std::optional<QXmppCallInviteElement::Jingle> jingle = std::nullopt,
+        std::optional<QVector<QXmppCallInviteElement::External>> external = std::nullopt);
```
lnjX, line 42: Maybe this should be private and only be accessed by the manager.

src/base/QXmppJingleData.h
```diff
@@ -657,6 +658,73 @@ class QXMPP_EXPORT QXmppJingleMessageInitiationElement
...
+        Reject,
+        Left
+    };
+
+    struct Jingle
+    {
+        QString sid;
+        std::optional<QString> jid;
+
+        bool operator==(const Jingle &other) const { return other.sid == sid && other.jid == jid; }
```
lnjX, line 678: Have you tried whether the implicit == opertor works?

src/base/QXmppJingleData.h
```diff
@@ -657,6 +658,73 @@ class QXMPP_EXPORT QXmppJingleMessageInitiationElement
...
+    void setExternal(std::optional<QVector<External>> external);
+
+    /// \cond
+    void parse(const QDomElement &element);
+    void toXml(QXmlStreamWriter *writer) const;
+    /// \endcond
+
+    static bool isCallInviteElement(const QDomElement &);
+    static QString callInviteElementTypeToString(Type type);
+    static std::optional<Type> stringToCallInviteElementType(const QString &typeStr);
```
lnjX, line 722: The type parse functions should probably rather be private.

src/base/QXmppJingleData.cpp
```diff
@@ -3224,3 +3224,319 @@ std::optional<QXmppJingleMessageInitiationElement::Type> QXmppJingleMessageIniti
 
     return std::nullopt;
 }
+
+class QXmppCallInviteElementPrivate : public QSharedData
+{
+public:
+    QXmppCallInviteElementPrivate() = default;
```
lnjX, line 3231: I think you can just remove this line :)

src/base/QXmppJingleData.cpp
```diff
@@ -3224,3 +3224,319 @@ std::optional<QXmppJingleMessageInitiationElement::Type> QXmppJingleMessageIniti
...
+    d->type = type.value();
+    d->id = element.attribute(QStringLiteral("id"));
+
+    switch (d->type) {
+    case Type::Invite:
+        d->audio = element.attribute(QStringLiteral("audio"), QStringLiteral("true")) == QStringLiteral("true") ? true : false;
+        d->video = element.attribute(QStringLiteral("video"), QStringLiteral("false")) == QStringLiteral("true") ? true : false;
+        // fall through
+    case Type::Accept: {
+        if (const auto &jingleElement = element.firstChildElement("jingle"); !jingleElement.isNull()) {
```
lnjX, line 3382: No const ref to temporary here please

src/base/QXmppJingleData.cpp
```diff
@@ -3224,3 +3224,319 @@ std::optional<QXmppJingleMessageInitiationElement::Type> QXmppJingleMessageIniti
...
+            if (!d->jingle) {
+                d->jingle = Jingle();
+            }
```
lnjX, line 3385: I think you can just always reset the jingle element when parsing.

src/client/QXmppCallInviteManager.h
lnjX: My thought here was that the call invites are only created by the manager and there's no API to create the call "manually". Was your plan that the users of QXmpp can "manually" create a `CallInvite()` with a manager and then calling `invite()` on the call invite?

If (only) `CallInviteManager::invite()` should be used to create invites, this and the `CallInvite(CallInviteManager *)` ctor should probably be made private (with friend class for the manager).

src/client/QXmppCallInviteManager.h
```diff
@@ -0,0 +1,123 @@
...
+    struct Retracted
+    {
+    };
+    struct Left
+    {
+    };
+
+    using Result = std::variant<Rejected, Retracted, Left, QXmppError>;
+
+    QXmppCallInvite(QXmppCallInviteManager *manager);
```
lnjX, line 35: should be `explicit` to prevent implicit conversion of CallInviteManager* into CallInvite.

src/base/QXmppJingleData.h
lnjX: ah, yes, sorry (with C++20 we could do ` = default;` but thats all).

src/base/QXmppJingleData.h
lnjX: It's just that if they're `public`, they are part of the public QXmpp API and need to stay compatible. APIs that are not needed by the users of QXmpp shouldn't be public, so we don't need to keep the compaibility of them.

Better make it `private:` or put it into `namespace QXmpp::Private {}`, if they are not needed by users of QXmpp.

src/base/QXmppJingleData.cpp
lnjX: yeah that line also doesn't seem to be necessary

src/base/QXmppJingleData.cpp
lnjX: It's not about thread-safety here.
* `firstChildElement()` returns by value (`QDomElement` not `const QDomElement &`).
* taking `const &` of the return value creates a reference to a temporary object (and I guess that's probably not what you wanted to do?)
* C++ says that the lifetimes of temporaries is extended in such cases though, so you won't get a dangling reference and the code works
* still i think that this is contra-intuitive and it looks like `firstChildElement()` would return a `const &`
* taking `const &` here is not faster, both variants generate the same binary
* there's nothing wrong with using `auto jingleElement = ...` (or am I overlooking something?)

src/base/QXmppJingleData.cpp
lnjX: Yes, what I tried to say is that you can replace `if (!d->jingle) { d->jingle = Jingle(); }` with just `d->jingle = Jingle();`. (inside the `!jingleElement.isNull()`)

lnjX: Yes and even if it is set, it's still fine to reset it here.

src/base/QXmppJingleData.cpp
```diff
@@ -3224,3 +3224,314 @@ std::optional<QXmppJingleMessageInitiationElement::Type> QXmppJingleMessageIniti
...
+    }
+
+    // write namespace and ID.
+    writer->writeDefaultNamespace(ns_call_invites);
+    helperToXmlAddAttribute(writer, QStringLiteral("id"), d->id);
+
+    // no need to go on for reject, retract and left tags.
+    if (d->type == Type::Reject || d->type == Type::Retract || d->type == Type::Left) {
+        return;
+    }
```
lnjX, line 3430: This code generates invalid XML, e.g. `<retract xmlns="urn:xmpp:call-invites:0"`.

Sorry that I didn't see this earlier. Please add a unit test.

lnjX:
```
$ make doc
Generate API documentation with Doxygen
qxmpp/src/client/QXmppCallInviteManager.h:33: warning: Member Result (typedef) of class QXmppCallInvite is not documented.
qxmpp/src/client/QXmppCallInviteManager.h:43: warning: Member invited() (function) of class QXmppCallInvite is not documented.
```

lnjX (approved):

lnjX:
@taiBsu It seems like this signal hasn't been documented:
```
qxmpp/src/client/QXmppCallInviteManager.h:43: warning: Member invited() (function) of class QXmppCallInvite is not documented.
```

PR #580, opened 2023-09-17
Don't search for your headers in your install dir, when building the lib

dfaure-kdab:
This leads to "not found" warnings with -Wmissing-include-dirs but worse, it will one day lead to build errors when having different APIs in the installed headers and locally (after making changes)

PR check list:
- [x] Document your code
- [x] Update `doc/doap.xml`
- [x] Add unit tests


lnjX (approved):
Thanks! :)

PR #582, opened 2023-10-07
MamManager: Inject encrypted messages for other managers

melvo:
QXmppMessageDeliveryManager can now send delivery receipts for encrypted messages as well.

PR check list:
- [x] Document your code
- [ ] Fix doxygen warnings (see log when building with `-DBUILD_DOCUMENTATION=ON`)
- [x] Update `doc/doap.xml`
- [ ] Add unit tests


lnjX (changes requested):
This is unfortunately no solution, because this way old, historical messages would be processed like new incoming messages.

PR #588, opened 2023-11-06
Replace '#ifndef QXMPP*_H' with '#pragma once'

melvo:
PR check list:
- [x] Document your code
- [x] Update `doc/doap.xml`
- [x] Add unit tests

### DIFF
--- a/src/base/QXmppArchiveIq.h
+++ b/src/base/QXmppArchiveIq.h
@@ -2,8 +2,7 @@
 //
 // SPDX-License-Identifier: LGPL-2.1-or-later
 
-#ifndef QXMPPARCHIVEIQ_H
-#define QXMPPARCHIVEIQ_H
+#pragma once
 
 #include "QXmppIq.h"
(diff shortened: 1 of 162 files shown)


lnjX:
I actually haven't replaced the header guards with `#pragma once` on purpose. `#pragma once` is afaik controversial and not every project using QXmpp might want to use it.

melvo:
> I actually haven't replaced the header guards with `#pragma once` on purpose. `#pragma once` is afaik controversial and not every project using QXmpp might want to use it.

@lnjX I thought using `#pragma once` would be the new way of doing it because you used it for the new files `QXmppPubSubBaseItem.h` and `QXmppBlockingManager.h`. What are the disadvantages?

lnjX:
Sorry, I must have used pragma once in the BlockingManager and PubSubBaseItem by accident.

Using pragma once probably isn't a big problem, but I still think it's better to use include guards in a library like this.
See e.g. https://stackoverflow.com/a/23699893

PR #594, opened 2023-11-28
Fix account creation/deletion

melvo:
PR check list:
- [ ] Document your code
- [ ] Add `\since QXmpp 1.X`, `QXMPP_EXPORT`
- [ ] Fix doxygen warnings (see log when building with `-DBUILD_DOCUMENTATION=ON`)
- [ ] Update `doc/doap.xml`
- [ ] Add unit tests
- [ ] Format the code: Run `clang-format -i src/<edited-file(s)> tests/<edited-file(s)>`


src/base/QXmppStanza.h
```diff
@@ -135,7 +135,8 @@ class QXMPP_EXPORT QXmppStanza : public QXmppNonza
             SubscriptionRequired,       ///< The requester needs to subscribe first.
             UndefinedCondition,         ///< An undefined condition was hit.
             UnexpectedRequest,          ///< The request was unexpected.
-            PolicyViolation             ///< The entity has violated a local server policy. \since QXmpp 1.3
+            PolicyViolation,            ///< The entity has violated a local server policy. \since QXmpp 1.3
+            EmailConfirmationRequired,  ///< The server requires the entitiy to confirm an email message that the server sent to the entity. \since QXmpp 1.6
```
lnjX, line 139: that's no stanza error, right? i don't think i can accept that.

melvo: The new QXmpp::StreamError is used now.

src/base/QXmppGlobal.h
```diff
@@ -178,7 +178,9 @@ struct Success { };
 struct Cancelled { };
 
 ///
-/// Unit struct indicating a timeout (or keep-alive) error.
+/// Unit struct indicating a client-side timeout (or keep-alive) error.
+///
+/// It occurs if the own server does not respond within a given timeout.
```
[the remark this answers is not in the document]
melvo: I suppose you suggested that because the error could occur when QXmpp is used as a server. Or are there other cases?We should maybe use the term `entity` instead of `peer` to be consistent since it is used at other places as well.

src/client/QXmppClient.h
```diff
@@ -335,6 +339,7 @@ private Q_SLOTS:
     friend class QXmppCarbonManagerV2;
     friend class QXmppRegistrationManager;
     friend class TestClient;
+    friend class QXmppRegistrationManager;
```
lnjX, line 342: duplicate

src/base/QXmppGlobal.h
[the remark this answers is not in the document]
lnjX: I used peer to keep it generic for other connections we've got, like STUN/TURN. I'm also fine with entity.

lnjX (approved):

lnjX (approved):

PR #633, opened 2024-06-01
Fix QXmppAccountMigrationManager issues

pasnox:
We can only request adding roster items one by one

PR check list:
- [X] Document your code
- [X] Add `\since QXmpp 1.X`, `QXMPP_EXPORT`
- [X] Update `doc/doap.xml`
- [X] Add unit tests
- [X] Format the code: Run `clang-format -i src/<edited-file(s)> tests/<edited-file(s)>`

### DIFF
--- a/src/client/QXmppRosterManager.cpp
+++ b/src/client/QXmppRosterManager.cpp
@@ -523,11 +523,36 @@ void QXmppRosterManager::onRegistered(QXmppClient *client)
 {
     // data import/export
     if (auto manager = client->findExtension<QXmppAccountMigrationManager>()) {
-        auto importData = [this](const RosterData &data) -> QXmppTask<std::variant<Success, QXmppError>> {
-            QXmppRosterIq iq;
-            iq.setItems(data.items);
-            iq.setType(QXmppIq::Set);
-            return this->client()->sendGenericIq(std::move(iq));
+        using Result = std::variant<Success, QXmppError>;
+        auto importData = [this](const RosterData &data) -> QXmppTask<Result> {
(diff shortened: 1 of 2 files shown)


src/client/QXmppVCardManager.cpp
```diff
@@ -87,9 +87,10 @@ QXmppTask<QXmppVCardManager::VCardIqResult> QXmppVCardManager::fetchVCard(const
 QXmppTask<QXmppVCardManager::Result> QXmppVCardManager::setVCard(const QXmppVCardIq &vCard)
 {
     auto vCardIq = vCard;
-    vCardIq.setTo(client()->configuration().jidBare());
+    vCardIq.setTo({});
     vCardIq.setFrom({});
     vCardIq.setType(QXmppIq::Set);
+    d->clientVCard = vCardIq;
     return client()->sendGenericIq(std::move(vCardIq));
 }
```
lnjX, line 95: Actually my intention was to deprecate the caching of the client VCard because we don't know when it changes and we can't say if it is up-to-date. The new `fetchVCard()` also doesn't set the client vcard. Though I must admit, the documentation could be better here. And the old functions should actually be marked as deprecated.

lnjX (commented):
The unit tests still fail, otherwise thanks for the fix :)

lnjX:
Ah and can you re-apply your commit onto `1.7`? So we can do a 1.7.1 release.

pasnox:
> The unit tests still fail, otherwise thanks for the fix :)

Fixed, remaining CI issues seems unrelated to this change.

lnjX (approved):

PR #639, opened 2024-06-29
QXmppAccountMigrationManager: Make public the extensions getter/setter

pasnox:
This is needed because without that change applications can not store and use app specific data. By example Kaidan export and import its roster settings which requires that all migration tasks finished before importing roster related data.

PR check list:
- [x] Document your code
- [x] Update `doc/doap.xml`
- [x] Add unit tests

### DIFF
--- a/src/client/QXmppAccountMigrationManager.cpp
+++ b/src/client/QXmppAccountMigrationManager.cpp
@@ -206,6 +206,9 @@ struct QXmppAccountMigrationManagerPrivate {
 ///
 /// Constructs an account migration manager.
 ///
+/// \note You would need the `QXmppClient(QXmppClient::NoExtensions, this)` approach to use this manager
+/// because it needs to be instantiated before others using it.
+///
 QXmppAccountMigrationManager::QXmppAccountMigrationManager()
     : d(std::make_unique<QXmppAccountMigrationManagerPrivate>())
 {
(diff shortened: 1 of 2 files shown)


src/client/QXmppAccountMigrationManager.cpp
```diff
@@ -206,6 +206,9 @@ struct QXmppAccountMigrationManagerPrivate {
 ///
 /// Constructs an account migration manager.
 ///
+/// \note You would need the `QXmppClient(QXmppClient::NoExtensions, this)` approach to use this manager
+/// because it needs to be instanciated before others using it.
```
lnjX, line 210:
```suggestion
/// because it needs to be instantiated before others using it.
```

src/client/QXmppAccountMigrationManager.h
```diff
@@ -57,13 +57,13 @@ class QXMPP_EXPORT QXmppExportData
...
+    const std::unordered_map<std::type_index, std::any> &extensions() const;
+    void setExtension(std::any value);
+
 private:
     friend class QXmppAccountMigrationManager;
     friend class tst_QXmppAccountMigrationManager;
 
-    const std::unordered_map<std::type_index, std::any> &extensions() const;
-    void setExtension(std::any value);
-
```
lnjX, line 66: Can you keep those functions private and add the following (public):

```
template<typename T>
std::optional<T> extensionData() const;

template<typename T>
void setExtensionData(T &&);
```

(those functions maybe need private helpers, not sure)

lnjX (approved):

lnjX:
Sorry for the delay. I just noticed QXmppExportData is missing any documentation. Would you like to fix that?

PR #657, opened 2024-10-06
Fix migration manager

pasnox:
PR check list:
- [x] Document your code
- [x] Update `doc/doap.xml`
- [x] Add unit tests

Fixed a bunch of issues in the QXmppAccountMigrationManager:

- Stabilized the unit tests moving from unordered to sorted keys nodes
- Handle correctly joining mix channels

### DIFF
--- a/src/base/Algorithms.h
+++ b/src/base/Algorithms.h
@@ -11,6 +11,16 @@
 
 namespace QXmpp::Private {
 
+template<typename T>
+constexpr bool HasShrinkToFit = requires(const T &t) {
+    t.shrink_to_fit();
+};
+
+template<typename T>
+constexpr bool HasSqueeze = requires(const T &t) {
+    t.squeeze();
+};
+
 template<typename OutputVector, typename InputVector, typename Converter>
 auto transform(const InputVector &input, Converter convert)
 {
@@ -20,6 +30,26 @@ auto transform(const InputVector &input, Converter convert)
     }
     for (const auto &value : input) {
         output.push_back(std::invoke(convert, value));
+    }
+    return output;
+}
+
+template<typename OutputVector, typename InputVector, typename Converter>
+auto transformFilter(const InputVector &input, Converter convert)
+{
+    OutputVector output;
+    if constexpr (std::ranges::sized_range<InputVector>) {
+        output.reserve(input.size());
+    }
+    for (const auto &value : input) {
+        if (const std::optional<std::decay_t<decltype(value)>> result = std::invoke(convert, value)) {
+            output.push_back(*result);
+        }
+    }
+    if constexpr (HasShrinkToFit<InputVector>) {
+        output.shrink_to_fit();
+    } else if constexpr (HasSqueeze<InputVector>) {
+        output.squeeze();
     }
     return output;
 }
--- a/src/client/QXmppAccountMigrationManager.cpp
+++ b/src/client/QXmppAccountMigrationManager.cpp
@@ -11,6 +11,7 @@
 #include "QXmppTask.h"
 #include "QXmppUtils_p.h"
 
+#include "Algorithms.h"
 #include "StringLiterals.h"
 
 #include <QDomElement>
@@ -30,6 +31,16 @@ struct XmlElementId {
     {
         return tagName == other.tagName && xmlns == other.xmlns;
     }
+
+    bool operator<(const XmlElementId &other) const {
+        const auto result = xmlns.compare(other.xmlns);
+
+        if (result == 0) {
+            return tagName.compare(other.tagName) < 0;
+        }
+
+        return result < 0;
+    }
 };
 
 #ifndef QXMPP_DOC
@@ -46,6 +57,12 @@ struct std::hash<XmlElementId> {
 
 using AnyParser = QXmppExportData::ExtensionParser<std::any>;
 using AnySerializer = QXmppExportData::ExtensionSerializer<std::any>;
+
+static std::unordered_map<std::type_index, XmlElementId> &accountDataMapping()
+{
+    thread_local static std::unordered_map<std::type_index, XmlElementId> registry;
+    return registry;
+}
 
 static std::unordered_map<XmlElementId, AnyParser> &accountDataParsers()
 {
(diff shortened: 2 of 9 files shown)


src/client/QXmppAccountMigrationManager.h
```diff
@@ -154,4 +156,12 @@ void QXmppAccountMigrationManager::unregisterExportData()
...
 }
 
+///
+/// \fn QXmppAccountMigrationManager::errorOccured(const QXmppError &error)
+///
+/// Emitted when an error occured during export or import.
+///
+/// \param error The occured error
+///
+
```
lnjX, line 166: should be in previous commit (or squashed)

src/client/QXmppAccountMigrationManager.cpp
```diff
@@ -161,7 +159,7 @@ struct QXmppAccountMigrationManagerPrivate {
         std::function<QXmppTask<Result<std::any>>()> exportFunction;
     };
 
-    std::unordered_map<std::type_index, ExtensionData> extensions;
+    std::map<std::type_index, ExtensionData> extensions;
```
lnjX, line 162: Using `map` instead of `unordered_map` should be avoided if possible. `map` is much slower than `unordered_map`

image

unordered_map does search, insert and delete in average O(1) and worst case O(log(n)), while map always takes O(log(n)) which makes a huge difference with large `n`.

Please rather fix the comparison in the unit tests.

pasnox: I'm well aware about the difference between map and unordered_map and the consequences of it.
This map will host at best ~20 entries, it's not as if it would make any difference at such low number. We would have to have 10k/15k at a minimum so that the difference starts to impact something.
Anyway, I don't know how I would allow std::unordered_map to have a stable seed/hash for unit tests, if you have idea or something i can do for that just tell me and I will happily revert/apply, else i will simply keep moving to map because as of now there is no downside to move to.

pasnox: I came up with a partial solution, see last commit and tell me if it's sufficient for you.

tests/TestClient.h
```diff
@@ -16,9 +16,10 @@ class TestClient : public QXmppClient
 {
     Q_OBJECT
 public:
-    TestClient(bool enableDebug = false)
+    TestClient(bool enableDebug = false, bool disableAutoReset = false)
```
melvo, line 19: That makes it easier to read/understand:
```suggestion
    TestClient(bool enableDebug = false, bool enableAutoReset = true)
```

src/client/QXmppRosterManager.cpp
```diff
@@ -27,7 +30,25 @@ using namespace QXmpp::Private;
 namespace QXmpp::Private {
 
 struct RosterData {
+    struct MixChannel {
+        QString channel;
+        QString nick;
+
+        void parse(const QDomElement &element) {
+            channel = element.attribute(u"channel"_s);
```
melvo, line 38:
```suggestion
            channel = element.attribute(u"jid"_s);
```

src/client/QXmppMovedManager.cpp
```diff
@@ -22,6 +22,12 @@
 using namespace QXmpp;
 using namespace QXmpp::Private;
 
+///
+/// \typedef QXmppMovedManager::Result
+///
+/// Empty result containing QXmpp::Success or a QXmppError
+///
+
```
melvo, line 30: I suggest to move that below the documentation for `QXmppMovedManager`.

melvo: Apart from that, how about simply using `QXmppClient::EmptyResult` which is already documented?

src/client/QXmppAccountMigrationManager.cpp
```diff
@@ -102,18 +102,28 @@ std::variant<QXmppExportData, QXmppError> QXmppExportData::fromDom(const QDomEle
...
+    }();
+
     writer->writeStartDocument();
     writer->writeStartElement(QSL65("account-data"));
     writer->writeDefaultNamespace(toString65(ns_qxmpp_export));
     writer->writeAttribute(QSL65("jid"), d->accountJid);
 
     const auto &serializers = accountDataSerializers();
-    for (const auto &[typeIndex, extension] : std::as_const(d->extensions)) {
+    for (const auto &typeIndex: sortedExtensionsKeys) {
```
lnjX, line 121:
```suggestion
    for (const auto &typeIndex : sortedExtensionsKeys) {
```

src/client/QXmppMixManager.cpp
```diff
@@ -34,6 +39,65 @@ class QXmppMixManagerPrivate
...
+            writeOptionalXmlAttribute(writer, u"nick", nick);
+            writer->writeEndElement();
+        }
+    };
+
+    QList<Item> items;
+
+    static std::variant<MixData, QXmppError> fromDom(const QDomElement &el)
+    {
+        if (el.tagName() != u"mix" && el.namespaceURI() != ns_qxmpp_export) {
```
lnjX, line 69:
```suggestion
        if (el.tagName() != u"mix" || el.namespaceURI() != ns_qxmpp_export) {
```

src/base/Algorithms.h
```diff
@@ -24,6 +24,21 @@ auto transform(const InputVector &input, Converter convert)
...
+    if constexpr (std::ranges::sized_range<InputVector>) {
+        output.reserve(input.size());
+    }
+    for (const auto &value : input) {
+        if (const std::optional<std::decay_t<decltype(value)>> result = std::invoke(convert, value)) {
+            output.push_back(*result);
+        }
+    }
+    return output;
+}
```
lnjX, line 40: Since you're using the capacity of the old vector (without filtering), I'd suggest to add a call to `output.shrink_to_fit()` in case many elements have been removed.

tests/qxmppmovedmanager/tst_qxmppmovedmanager.cpp
lnjX, line 1: I fixed this now on master by using u"" also on the second part:
https://github.com/qxmpp-project/qxmpp/commit/85420824ab96922647c98a2cc2d0fe97a4f56183

lnjX:
[QXmppMovedManager: Fix doxygen missing documentation](https://github.com/qxmpp-project/qxmpp/pull/657/commits/2c06651baf891939af8a30a28bd8b079939d0381)

I fixed this on master now without using QXmppClient::EmptyResult.

The alias `QXmppClient::EmptyResult` should not be used in other classes as QXmppClient isn't really the right place for a global alias and I'd also like to avoid including QXmppClient.h in every other file.

lnjX:
>  Util: Add more helpers 

Can you change the commit description so that it properly describes the changes (not just "more utils"), e.g. `tests: util: Add xmlToFormattedByteArray() func`.

lnjX:
> QXmppRosterManager: Handle better mix roster entries 

Please include that this is only about data import/export in the commit description. And in the long description you could note what is actually fixed.

lnjX (approved):
apart from last minor comments looks good :)

PR #670, opened 2024-11-18
MixManager: Fix behavior for updating node subscriptions

melvo:
If no nodes were specified to be unsubscribed from for updateSubscriptions(), all enums were added to be unsubscribed from. That resulted in the same nodes being specified to be subcribed to and to be unsubscribed from.

PR check list:
- [x] Document your code
- [x] Update `doc/doap.xml`
- [x] Add unit tests

### DIFF
--- a/tests/qxmppmixmanager/tst_qxmppmixmanager.cpp
+++ b/tests/qxmppmixmanager/tst_qxmppmixmanager.cpp
@@ -1163,11 +1163,84 @@ void tst_QXmppMixManager::testUpdateSubscriptions()
     auto &client = tester.client;
     auto manager = tester.manager;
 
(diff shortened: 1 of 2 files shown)


tests/qxmppmixmanager/tst_qxmppmixmanager.cpp
```diff
@@ -1163,11 +1163,84 @@ void tst_QXmppMixManager::testUpdateSubscriptions()
     auto &client = tester.client;
     auto manager = tester.manager;
 
+    auto defaultParametersCall = [&client, manager]() {
+        return manager->updateSubscriptions(QStringLiteral("coven@mix.shakespeare.example"));
```
lnjX, line 1167: s/QStringLiteral/u""_s/

PR #676, opened 2025-01-22
Qt 6: Fix crash when sending encrypted files

pasnox:
m_cipher->ok() is to be called only after an explicit `update()` or `final()` call.
This changes fix the crash when attempting to upload encrypted files.

PR check list:
- [ ] Document your code
- [ ] Add `\since QXmpp 1.X`, `QXMPP_EXPORT`
- [ ] Update `doc/doap.xml`
- [ ] Add unit tests
- [ ] Format the code: Run `clang-format -i src/<edited-file(s)> tests/<edited-file(s)>`

### DIFF
--- a/src/client/QXmppFileEncryption.cpp
+++ b/src/client/QXmppFileEncryption.cpp
@@ -152,7 +152,6 @@ EncryptionDevice::EncryptionDevice(std::unique_ptr<QIODevice> input,
     setOpenMode(m_input->openMode() & QIODevice::ReadOnly);
 
     Q_ASSERT(m_cipher->validKeyLength(int(key.length())));
-    Q_ASSERT(m_cipher->ok());
 }
 
 EncryptionDevice::~EncryptionDevice() = default;
@@ -268,7 +267,6 @@ DecryptionDevice::DecryptionDevice(std::unique_ptr<QIODevice> input,
     setOpenMode(m_output->openMode() & QIODevice::WriteOnly);
 
     Q_ASSERT(m_cipher->validKeyLength(int(key.length())));
-    Q_ASSERT(m_cipher->ok());
 }
 
 DecryptionDevice::~DecryptionDevice() = default;


src/client/QXmppFileEncryption.cpp
```diff
@@ -1,4 +1,5 @@
 // SPDX-FileCopyrightText: 2022 Linus Jahn <lnj@kaidan.im>
+// SPDX-FileCopyrightText: 2025 Filipe Azevedo <pasnox@gmail.com>
```
lnjX, line 2: This is only done when a relevant amount of new code is added.

lnjX (approved):

src/client/QXmppFileEncryption.cpp
melvo: > Also, my PRs were not enabling edits by maintainers, i'm a bit surprised GH allowed you to write in my branch in such case.

Our procedure is usually to allow the other devs to push to our own PR branches in order to fix something directly without too much friction.

> Not that i'm looking for any credit on any sort, but i'm a bit disappointed on the stance you take here.
So for you the hours passed debugging your lib does not deserve credit ?
Any code change, being a simple comment typo to one line code change deserve credit in general opinions.

I think there are different opinions on when to add someone to the license header. One could argue that any change should result in an addition. But at least for QXmpp and Kaidan, we agreed on doing that only for major changes on the code level because only that means to have a relevant portion of authorship claims. How much time a contributor needed to fix something is not known to the maintainer. Thus, it cannot be taken into account following the mentioned agreement even if I can understand that it seems to be unfair.

lnjX: Sorry I first read this comment today. I don't want to not give you credit. The copyright header is for the legal stuff and removing lines or doing simple changes doesn't affect authorship. For attribution, I like to give credit in the release notes (See https://github.com/qxmpp-project/qxmpp/releases/tag/v1.9.3 ) and in the CHANGELOG. And some people may also look at the commits and PRs themselves.

